PR #2, opened 2015-01-05
updates

stefanpenner:
* prefer official 6to5 broccoli filter
* lazily load broccoli-6to5-transpiler to save 250ms from ember-cli boot.
* add node_modules to gitignore


gordonkristan:
I didn't know there was a broccoli plugin for 6to5, I just kind of dove in and wrote my own. :) Thanks for pointing that out.

stefanpenner:
:+1:

PR #6, opened 2015-01-15
Support including the 6to5 polyfill when requested.

dfreeman:
6to5 relies on a custom polyfill for [some features](https://6to5.org/docs/caveats/) (generators, symbols for iteration, etc). This PR adds an `includePolyfill` flag to incorporate that file into the build.

Since the polyfill patches built-in objects, I left the inclusion off by default, but I could also see an argument for making the behavior opt-out instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,7 @@
 var checker   = require('ember-cli-version-checker');
+var clone     = require('clone');
+var path      = require('path');
+var resolve   = require('resolve');
 
 module.exports = {
   name: 'ember-cli-babel',
@@ -14,9 +17,28 @@ module.exports = {
       name: 'ember-cli-babel',
       ext: 'js',
       toTree: function(tree) {
-        return require('broccoli-babel-transpiler')(tree, getOptions(addon));
+        return require('broccoli-babel-transpiler')(tree, getBabelOptions(addon));
       }
     });
+  },
+
+  shouldIncludePolyfill: function() {
+    var options = getAddonOptions(this);
+    return options.includePolyfill === true;
+  },
+
+  importPolyfill: function(app) {
+    app.import('vendor/browser-polyfill.js');
+  },
+
+  treeFor: function(name) {
+    if (name !== 'vendor') { return; }
+
+    // Find babel-core's browser polyfill and use its directory as our vendor tree
+    var transpilerRoot = path.dirname(resolve.sync('broccoli-babel-transpiler'));
+    var polyfillDir = path.dirname(resolve.sync('babel-core/browser-polyfill', { basedir: transpilerRoot }));
+
+    return this.treeGenerator(polyfillDir);
   },
 
   included: function(app) {
@@ -26,12 +48,21 @@ module.exports = {
     if (this.shouldSetupRegistryInIncluded()) {
       this.setupPreprocessorRegistry('parent', app.registry);
     }
+
+    if (this.shouldIncludePolyfill()) {
+      this.importPolyfill(app);
+    }
   }
 };
 
-function getOptions(addonContext) {
-  var baseOptions = (addonContext.parent && addonContext.parent.options) || (addonContext.app && addonContext.app.options),
-      options = baseOptions && baseOptions['babel'] || {};
+function getAddonOptions(addonContext) {
+  var baseOptions = (addonContext.parent && addonContext.parent.options) || (addonContext.app && addonContext.app.options);
+  return baseOptions && baseOptions.babel || {};
+}
+
+function getBabelOptions(addonContext) {
+  var options = clone(getAddonOptions(addonContext));
+
   // Ensure modules aren't compiled unless explicitly set to compile
   options.blacklist = options.blacklist || ['es6.modules'];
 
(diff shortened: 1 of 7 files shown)


stefanpenner:
will review friday night or saturday morning..
Thanks!

gordonkristan:
The only thing that makes me a little weary is that we're relying on a dependency of a dependency. We recently had our build break at work when a dependency of a dependency introduced a bug in a patch release. Is it possible to make 6to5 a direct dependency of this, ensuring that the polyfill won't change unless we explicitly upgrade the version?

dfreeman:
@gordonkristan Reaching into the files for the transitive dep definitely felt off to me. The problem (I think) with pulling in `6to5-core` directly, though, is there's the potential for it to end up resolving to a different version than the one the transpiler uses. If that happens, and 6to5 changes the polyfill between those two versions, the built artifact could end up with subtly (or not-so-subtly) broken.

That said, I don't have a super deep understanding of how NPM's dependency resolution works, so there may be a way to avoid the version mismatch problem. Hopefully @stefanpenner (who has a lot more experience working with this kind of thing) can suggest the right way forward.

index.js
```diff
@@ -1,22 +1,43 @@
...
+    name: 'ember-cli-6to5',
+    ext: 'js',
+    toTree: function(tree) {
+      return require('broccoli-6to5-transpiler')(tree, options);
+    }
+  });
+}
+
+function includePolyfill(app) {
+  app.import('vendor/6to5-core/browser-polyfill.js');
```
stefanpenner: i dont think we can actually even rely on this location. NPM dedupe strategy can hoist if needed, likely using resolve to derive the correct path is needed. I have some thoughts, will noodle on it more this evening.

dfreeman: Good point; I hadn't thought about the complications from deduping. Seems like this would also impact [ember-cli-ic-ajax's implementation](https://github.com/rwjblue/ember-cli-ic-ajax/blob/master/index.js#L9), which was mentioned in some thread as the canonical way to handle this kind of import. The likelihood of ic-ajax getting hoisted is probably a lot lower, though, given that there aren't many other packages out there that depend on it.

stefanpenner: > The likelihood of ic-ajax getting hoisted is probably a lot lower, though, given that there aren't many other packages out there that depend on it.

just wait until we start supporting nested-addons... I started that effort this week... talk about pandoras box

index.js
```diff
@@ -1,22 +1,43 @@
+var path = require('path');
+
 module.exports = {
   name: 'ember-cli-6to5',
+
+  treeFor: function(name) {
+    if (name !== 'vendor') { return; }
+
+    // Use the transpiler's node_modules directory the vendor tree for this addon
+    return this.treeGenerator(path.join(__dirname, 'node_modules', 'broccoli-6to5-transpiler', 'node_modules'));
```
stefanpenner: this will need to use resolve, or the instability of npm's deduping may change this path out from under you.

dfreeman: Totally dropped the ball on this -- sorry! I made the change locally but never followed up and pushed, and now there are a bunch of upstream changes. I'll deal with the fallout from that and follow up shortly.

dfreeman:
@stefanpenner I'm running into issues grabbing the options hash, as mentioned in [this comment](https://github.com/babel/ember-cli-babel/commit/349a844d2bd044e1778ce2c0c4ebf3c2f1c437e2#commitcomment-10021664). Not positive, but I don't think the options are actually being correctly passed through in the non-nested case now?

Is the project environment the expected location for defining addon config moving forward? If so, is something like [how ember-cli-sass handles it](https://github.com/aexmachina/ember-cli-sass/blob/master/index.js#L31) how we should be looking for configuration in a backwards-compatible manner?

dfreeman:
This is back up to date with `master` and making use of `resolve` to track down the polypill now.

index.js
```diff
@@ -13,28 +16,60 @@ module.exports = {
...
     registry.add('js', plugin);
   },
 
+  shouldIncludePolyfill: function() {
+    var options = getAddonOptions(this);
+    return options.includePolyfill === true;
+  },
+
+  importPolyfill: function(app) {
+    app.import('vendor/browser-polyfill.js');
```
pgrippi: Should `{ prepend: true }` be added to this import statement so that the polyfill is included before Ember itself?

dfreeman: Probably wouldn't hurt, but is it likely to make an observable difference? Obviously Ember can't rely on ES6 APIs being available, but I'm not sure whether it behaves any differently in their presence.

jayphelps:
:cry:

hhff:
Hi all, any reason why this is hanging?  LGTM

stefanpenner:
@hhff i think we just need a rebase, and we can merge this.

dfreeman:
@stefanpenner Rebased again, and tweaked the acceptance test just to verify the polyfill makes it in when the flag is set.

stefanpenner:
FYI, we forgot to filter correctly. This resulted in us accidentally pulling in all of babel-core.... https://github.com/babel/ember-cli-babel/pull/48

Issue #18, opened 2015-02-25
compileModules: true doesn't work


mitchellmebane:
I spun up a new `ember-cli` app, installed `ember-cli-babel`, and added `'babel': { compileModules: true }` in my `Brocfile.js`.  When I run `ember server`, I get:

```
TypeError: Object #<Object> has no method 'indexOf'
    at getOptions (/var/node/testapp/node_modules/ember-cli-babel/index.js:28:40)
    at Class.module.exports.included (/var/node/testapp/node_modules/ember-cli-babel/index.js:6:19)
    at EmberApp.<anonymous> (/var/node/testapp/node_modules/ember-cli/lib/broccoli/ember-app.js:277:15)
    at Array.filter (native)
    at EmberApp._notifyAddonIncluded (/var/node/testapp/node_modules/ember-cli/lib/broccoli/ember-app.js:272:45)
    at new EmberApp (/var/node/testapp/node_modules/ember-cli/lib/broccoli/ember-app.js:236:8)
    at Object.<anonymous> (/var/node/testapp/Brocfile.js:5:11)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
```

It looks to me like this line in index.js is the culprit:

    options.blacklist.splice(options.indexOf('es6.modules'), 1);

I think it should be:

    options.blacklist.splice(options.blacklist.indexOf('es6.modules'), 1);

---

Something is still missing, though, because changing that leads to more errors:

```
ReferenceError: Unknown option: compileModules
    at File.normalizeOptions
```

mitchellmebane:
I added `delete options.compileModules`, and I *think* it's working properly now.  I guess I need to actually proceed with learning Ember now...

```javascript
var blacklistOptionIdx = options.blacklist.indexOf('es6.modules');
if (options.compileModules === true) {
  if (blacklistOptionIdx >= 0) {
    options.blacklist.splice(blacklistOptionIdx, 1);
  }
  delete options.compileModules;
} else {
  if (blacklistOptionIdx < 0) {
    options.blacklist.push('es6.modules');
  }
}
```

stefanpenner:
why not let ember-cli continue to handle the module transpilation?

mitchellmebane:
No real reason.

stefanpenner:
our plan is to continue to utilize esperanto which will allow folding and other interesting bundled outputs. Currently babel does not aim to support this.

[...]

PR #20, opened 2015-02-25
Fix for bug mentioned in #18

gordonkristan:
I don't even know if we really want to keep this option around any more, but if we do we'll have to fix it since #18 points out that it's currently broken. This should do the trick, but I wouldn't mind somebody else looking at it.


stefanpenner:
LGTM

PR #26, opened 2015-03-09
Resolves #25 - handle non-nested addon options

### LINKED ISSUE #25: not picking up custom 'babel' options in 0.2.0 ember-cli
I am trying to prevent the default `compact` behavior of babel

I set 
```
new EmberApp({
    babel: {
        compact: false
    },
```
that option is never seen in the invocation of babel

I'm logging this.parent.options in setupPreprocessorRegistry and it's undefined.

logging this while I look for correct solution

jschilli:
This works but feels a tad dirty.

`setupPreprocessorRegistry` is called super early in the process so I capture the app in included - which does make app.options available before the tree is processed.

I'm open to refinements and suggestions here if there is a cleaner path.

Seems like an addon method to determine if its nested or not would be a nice addition

cc: @lukemelia @stefanpenner

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,13 +8,12 @@ module.exports = {
   },
 
   setupPreprocessorRegistry: function(type, registry) {
-    var options = getOptions(this.parent && this.parent.options && this.parent.options['babel']);
-
+    var addon = this;
     var plugin = {
       name: 'ember-cli-babel',
       ext: 'js',
       toTree: function(tree) {
-        return require('broccoli-babel-transpiler')(tree, options);
+        return require('broccoli-babel-transpiler')(tree, getOptions(addon));
       }
     };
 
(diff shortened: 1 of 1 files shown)


rwjblue:
`this.parent` is intended to be the project for non-nested addons (aka direct project dependencies), and the parent addon in the nested addon case.  Unfortunately, the current config story kinda sucks right now and there is no way to provide build time configuration to the `project` instance without polluting the runtime config (in `index.html`'s `<meta>`).

This fix is good for now, but we need to fix the config story to get a path forward to avoiding the fallback lookup added here.

**tldr;** :+1:

index.js
```diff
@@ -8,13 +8,12 @@ module.exports = {
   },
 
   setupPreprocessorRegistry: function(type, registry) {
-    var options = getOptions(this.parent && this.parent.options && this.parent.options['babel']);
-
+    var that = this;
```
stefanpenner: lets call this `addon` to be clear what it is.

jschilli:
not sure what happened with that build

stefanpenner:
rebuilding....

jschilli:
I'm going to push a new branch to see if this will build cleanly - not sure what happened - my force push to my branch (to squash) may have bungled it

PR #34, opened 2015-03-28
Limit babelification to .es6 extensions only - do not merge

jschilli:
I have a cleaner fix to this issue at https://github.com/babel/broccoli-babel-transpiler/pull/15

We have a very large project with limited (currently) use of es6 features beyond modules.

The performance delta is documented in: https://github.com/ember-cli/ember-cli/issues/3637#issuecomment-85745165

I'm happy to make the change (configurable extensions to babelize with .js as the default) in either project.

would like feedback as to preference for this addon or broccoli-babel-transpiler

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -3,7 +3,11 @@
 
 var EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
-var app = new EmberAddon();
+var app = new EmberAddon({
+  babel: {
+    validExtensions:['js','es6'] // Support both .es6 and .js extensions for babel
+  }
+});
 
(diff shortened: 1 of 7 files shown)


jschilli:
I went ahead and made this full blown and configurable.  I still think the broccoli-babel-transpiler version is cleaner as we're already cloning options there

stefanpenner:
> I went ahead and made this full blown and configurable. I still think the broccoli-babel-transpiler version is cleaner as we're already cloning options there

yes I think that version is cleaner. I left some coments there

jschilli:
closing in favor of 
https://github.com/babel/broccoli-babel-transpiler/pull/15

PR #42, opened 2015-05-20
Possibility to add babel plugins via registry

pangratz:
This allows Ember-CLI addons to add custom babel transformations to the
ember-cli-babel pipeline by registering a plugin using the
`babel-plugin` type:

``` javascript
// my-addon/index.js
module.exports = {
  name: 'my-addon-using-babel-transformer',

  setupPreprocessorRegistry: function(type, registry) {
    var BabelTransformerPlugin = require('./babel-transformer-plugin');

    registry.add('babel-plugin', {
      name: 'my-addon-using-babel-transformer',
      plugin: BabelTransformerPlugin
    });
  }
}

// my-addon/babel-transformer-plugin.js
module.exports = function(babel) {
  var t = babel.types;
  return new babel.Transformer('troll-transformer', {
    Literal: function(node) {
      if (node.value === true) {
        return t.literal(false);
      } else if (node.value === false) {
        return t.literal(true);
      }
    }
  });
}
```

---

I will add a section to the README if this is accepted.

This addresses pangratz/ember-cli-htmlbars-inline-precompile#7

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,10 @@ module.exports = {
   setupPreprocessorRegistry: function(type, registry) {
     var addon = this;
 
+    if (type === 'parent') {
+      this.parentRegistry = registry;
+    }
+
     registry.add('js', {
       name: 'ember-cli-babel',
       ext: 'js',
@@ -89,6 +93,15 @@ function getBabelOptions(addonContext) {
       options.blacklist.push('es6.modules');
     }
   }
+
+  // get all babel-plugins in the registry
+  var pluginWrappers = addonContext.parentRegistry.load('babel-plugin');
+  var babelPlugins = pluginWrappers.map(function(wrapper) {
+    return wrapper.plugin;
+  });
+
+  // add babel-plugins from the registry to the options.plugins
+  options.plugins = (options.plugins || []).concat(babelPlugins);
 
   // Ember-CLI inserts its own 'use strict' directive
   options.blacklist.push('useStrict');
--- a/index.js
+++ b/index.js
@@ -96,7 +96,24 @@ function getBabelOptions(addonContext) {
 
   // get all babel-plugins in the registry
   var pluginWrappers = addonContext.parentRegistry.load('babel-plugin');
-  var babelPlugins = pluginWrappers.map(function(wrapper) {
+
+  // currently only a specific set of babel-plugin's is added to the pipeline:
+  // the reason is that if the ember-cli addon "ember-cli-my-addon" needs
+  // "babel-plugin-1" and an app uses the "ember-cli-my-addon" but doesn't
+  // list the "babel-plugin-1" in its package.json, the "ember-cli-my-addon" is
+  // not correctly pre-processed for the app, since "babel-plugin-1" is not
+  // available in the build pipeline
+  //
+  // as a current workaround for this, the allowed plugins are whitelisted
+  // here, until a general solution exists. see
+  // https://github.com/babel/ember-cli-babel/pull/42 for further context.
+  var allowedBabelPlugins = ['ember-cli-htmlbars-inline-precompile'];
+  var babelPlugins = pluginWrappers.filter(function(wrapper) {
+    // only allow whitelisted plugins
+    return allowedBabelPlugins.indexOf(wrapper.name) !== -1;
+  });
+
+  babelPlugins = babelPlugins.map(function(wrapper) {
     return wrapper.plugin;
   });
 
--- a/README.md
+++ b/README.md
@@ -38,6 +38,35 @@ features that require the polyfill to work.
 
 To include it in your app, pass `includePolyfill: true` in your `babel` options.
 
+### Adding custom babel plugins
+
+Custom [babel plugins](https://babeljs.io/docs/usage/plugins/) can be added to
+the transformation pipeline via Ember-CLI addons. The specific transformations
+needs to be added to the registry using `babel-plugin` as type:
+
+```js
+// my-babel-transformation-addon/index.js
+module.exports = {
+  name: 'my-babel-transformation-addon',
+
+  setupPreprocessorRegistry: function(type, registry) {
+    var MyBabelTransformation = require('./my-babel-transformation');
+
+    registry.add('babel-plugin', {
+      name: 'my-babel-transformation',
+      plugin: MyBabelTransformation
+    });
+  }
+}
+
+// my-babel-transformation-addon/my-babel-transformation.js
+module.exports = function(babel) {
+  return new babel.Transformer('my-babel-transformation', {
+    // see http://babeljs.io/docs/usage/plugins/manipulation/ for examples
+  });
+}
+```
+
 ### About Modules
 
 Ember-CLI uses its own ES6 module transpiler for the custom Ember resolver that it uses. Because of that,
--- a/lib/.jshintrc
+++ b/lib/.jshintrc
@@ -1,0 +1,4 @@
+{
+  "node": true,
+  "browser": false
+}
--- a/lib/answer-to-the-universe/index.js
+++ b/lib/answer-to-the-universe/index.js
@@ -1,0 +1,16 @@
+module.exports = {
+  name: 'answer-to-the-universe',
+
+  isDevelopingAddon: function() {
+    return true;
+  },
+
+  setupPreprocessorRegistry: function(type, registry) {
+    var ProvideAnswerTransformation = require('./provide-answer-transformation');
+
+    registry.add('babel-plugin', {
+      name: 'answer-to-the-universe',
+      plugin: ProvideAnswerTransformation
+    });
+  }
+};
--- a/lib/answer-to-the-universe/package.json
+++ b/lib/answer-to-the-universe/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "answer-to-the-universe",
+  "keywords": [
+    "ember-addon"
+  ]
+}
--- a/lib/answer-to-the-universe/provide-answer-transformation.js
+++ b/lib/answer-to-the-universe/provide-answer-transformation.js
@@ -1,0 +1,11 @@
+module.exports = function(babel) {
+  var t = babel.types;
+
+  return new babel.Transformer('answer-to-the-universe', {
+    Literal: function(node) {
+      if (node.value === 'THE-answer') {
+        return t.literal(42);
+      }
+    }
+  });
+};
--- a/lib/ember-cli-htmlbars-inline-precompile-mock/index.js
+++ b/lib/ember-cli-htmlbars-inline-precompile-mock/index.js
@@ -1,0 +1,16 @@
+module.exports = {
+  name: 'ember-cli-htmlbars-inline-precompile',
+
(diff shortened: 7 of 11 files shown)


stefanpenner:
mind adding a test of some kind?

rwjblue:
@pangratz - This looks good.

index.js
```diff
@@ -52,6 +52,15 @@ function getOptions(addonContext) {
     }
   }
 
+  // add babel-plugins from the registry to the options.plugins
+  var pluginWrappers = addonContext.registry.load('babel-plugin');
+  var babelPlugins = pluginWrappers.map(function(wrapper) {
+    return wrapper.plugin;
```
pangratz: Should the property under which the transformer is looked up be changed to `transformer`? I copied this from the `ember-cli-htmlbars` addon, where AST plugins are added like this ...

rwjblue: I think they are called plugins in babel speak also, right?

pangratz: https://babeljs.io/docs/usage/plugins/ yep, I think it's ok then

pangratz:
> mind adding a test of some kind?

done

pangratz:
Do NOT merge yet, I have to check if the plugin lookup works as expected ...

index.js
```diff
@@ -52,6 +52,15 @@ function getOptions(addonContext) {
     }
   }
 
+  // add babel-plugins from the registry to the options.plugins
+  var pluginWrappers = addonContext.registry.load('babel-plugin');
```
pangratz: Am I right that using `addonContext.registry` only looks up plugins within this addon, aka `in-repo-addon`s? Because if I'm using this in `ember-cli-htmlbars-inline-precompile' and register a `babel-plugin` it won't be picked up. Using `addonContext.app.registry` instead correctly picks up the plugins registered in other addons. @rwjblue can you confirm this?

rwjblue: You want `addonContext.parent.registry` (`addonContext.app` is not always available).

pangratz: Hmm, using `addonContext.parent.registry.load('babel-plugin') gives me a `Cannot read property 'load' of undefined`.

pangratz: Would something like currently used to get [the options](https://github.com/babel/ember-cli-babel/blob/5bde7459ef00811a13cf64c28660bd0f33f81a34/index.js#L33) be ok?

```
var registry = (addonContext.parent && addonContext.parent.registry) || (addonContext.app && addonContext.app.registry);
var pluginWrappers = registry.load('babel-plugin');
...
```

rwjblue: I would prefer that we stash the registry instance we are passed for `parent` in the `setupPreprocessorRegistry` hook.

Something like:

```javascript
setupPreprocessorRegistry: function(type, registry) {
  if (type === 'parent') {
    this.parentRegistry = registry;
  }
}
```

Then you should be able to call `addonContext.parentRegistry.load` without issue (or hacks).

pangratz: Nice!

pangratz:
Ok, so from my side this is good to go. I tested this within another addon and adding `babel-plugin`s to the transformation pipeline works as expected :raised_hands:

rwjblue:
WOOT

rwjblue:
@stefanpenner - Any additional feedback / tweaks?

stefanpenner:
I'll take a thorough look today.

rwjblue:
@stefanpenner - Any concerns after your review?

stefanpenner:
have not had time to review.

pangratz:
Since babel is moving so fast and features are added on an hourly basis, should there be some kind of check if the currently installed babel version satisfies the needs of the specific plugin? Something like:

``` js
registry.add('babel-plugin', {
  name: 'my-addon-using-babel-transformer',
  plugin: BabelTransformerPlugin,
  babelVersion: '5.2.10'
});
```

---

Currently *all* registered plugins of `babel-plugin` type are added to the pipeline. This means that also the `answer-to-the-universe` plugin of this PR is added (I will check if this is true in a sample app later). *If* there is a string like `THE-answer` in the app, it will be replaced by `42`...

rwjblue:
Having the allowed babel version range be required sounds nice.  We could then use the `semver` package to ensure we are in compliance of that range (and if not warn + skip).

Though, I think that could be done in a follow-up PR if @stefanpenner has had a chance to review this and is OK with it landing...

pangratz:
> Having the allowed babel version range be required sounds nice. We could then use the semver package to ensure we are in compliance of that range (and if not warn + skip).

:+1:

rwjblue:
@stefanpenner - Ping

rwjblue:
I spoke with @stefanpenner about this at our last face to face meeting.  We need an example app that has addons with differing versions of ember-cli-babel and using different plugins to make sure everything is handled correctly. Double checking that things are not shared is the main blocker for this PR.

@pangratz - Do you think that you might be able to put together a demo app like that?

pangratz:
I'm on it

pangratz:
@rwjblue @stefanpenner I created a sample app at https://github.com/pangratz/ember-cli-babel-plugins-sample, which uses the `ember-cli-babel` [version of this PR](https://github.com/pangratz/ember-cli-babel-plugins-sample/commit/046a582982908ea9d333b960ef949bd574890abe). Two addons are installed:

- [babel-plugin-demo1](https://github.com/pangratz/ember-cli-babel-plugin-demo1) which also uses the [`ember-cli-babel` version of this PR](https://github.com/pangratz/ember-cli-babel-plugin-demo1/blob/c49752ab00e11ffd22fd2bab6e3f68866acf88e8/package.json#L41)
- [babel-plugin-demo2](https://github.com/pangratz/ember-cli-babel-plugin-demo2) which uses [a different version of `ember-cli-babel` of this PR](https://github.com/pangratz/ember-cli-babel-plugin-demo2/blob/8cbb2cce098dc81096be758e008d84711579484c/package.json#L41) [(an empty commit has been added)](https://github.com/pangratz/ember-cli-babel/commit/e5b6fd386862a1f44830de9b2008b3dc4b9bf2c7)

The sample app has [a test case](https://github.com/pangratz/ember-cli-babel-plugins-sample/blob/f69a1aefaeae29cb51ab235693345440940f5438/tests/unit/babel-plugin-test.js), which can only succeed when both plugins work (the first plugin [replaces `babel-plugin-demo1` with `demo1`](https://github.com/pangratz/ember-cli-babel-plugins-sample/blob/f69a1aefaeae29cb51ab235693345440940f5438/tests/unit/babel-plugin-test.js#L6); and the second replaces [`babel-plugin-demo2` with `demo2`](https://github.com/pangratz/ember-cli-babel-plugins-sample/blob/f69a1aefaeae29cb51ab235693345440940f5438/tests/unit/babel-plugin-test.js#L10)).

As the [Travis run](https://travis-ci.org/pangratz/ember-cli-babel-plugins-sample/builds/66563470) shows, everything works as expected...

stefanpenner:
@pangratz this is a great start, thank you.

The test case should at the very least, include another level of addons.

`app -> my-cool-addon -> some-babel-plugin`

`some-babel-plugin` should affect the `addon/` directory of `my-cool-addon` but not the host application.

pangratz:
> my-cool-addon may also choose to op into different babel options.

Do you have an idea how to check that options enabled in `my-cool-addon` are not available/enabled in the host application?

stefanpenner:
> Do you have an idea how to check that options enabled in my-cool-addon are not available/enabled in the host application?

one idea, is to disable something like arrow functions in the parent, but enable them in one of the children.

pangratz:
:flushed: I think I need help on how to set/overwrite the babel options from inside an addon. Where is the right place for that?

stefanpenner:
> I think I need help on how to set/overwrite the ember-cli-babel options from inside an addon. Where is the right place for that?

I am pressed for time today, if you can put together an app that has the 3 layers described. (even if the options stuff doesn't work yet) I will attempt to dedicate some hours this weekend to it :)

pangratz:
Ok, I gave it a try and removed the `babel-plugin-demo2` dependency from the sample app and created a `babel-demo-addon` which depends on the `babel-plugin-demo2` addon.

The `babel-demo-addon` provides a `{{my-component}}`, which has a `leProperty` which is replaced via the `babel-plugin-demo2`: https://github.com/pangratz/babel-demo-addon/commit/b6cb25a5476778b13cd9ac7a8f2ab14a4f51c2fc

The sample app uses `babel-demo-addon` to check if the `babel-plugin-demo2` is correctly setup for within this addon, but not for the sample app itself, which it seems that currently **doesn't work as expected** ([failing travis run](https://travis-ci.org/pangratz/ember-cli-babel-plugins-sample/builds/66651258) for https://github.com/pangratz/ember-cli-babel-plugins-sample/commit/ae80ab7ae040205b72ae3e3c19557723e94fdfef).

---

I hope I understood you correctly regarding setup up the hierarchy and gave you both access to the relevant repos so you can check around if you have time ...

stefanpenner:
@pangratz I will confirm, but I believe this is the scenario I was concerned we missed.

stefanpenner:
there is a path, where we merge this in early. ANd explicity disallow the addon case, or hard-wire some of the common plugins we want.

rwjblue:
Let's just have a separate PR explicitly add the inline precompiler for now. I want to unblock the component integration test situation....

stefanpenner:
> I want to unblock the component integration test situation....

:+1:

pangratz:
> Let's just have a separate PR explicitly add the inline precompiler for now

Will create such a PR tomorrow ...

pangratz:
Instead of creating a new PR I've updated this one by only adding whitelisted `babel-plugin`s to the transformation pipeline (currently only `ember-cli-htmlbars-inline-precompile`)...

stefanpenner:
@rwjblue r?

pangratz:
@rwjblue did you have time to think about the proposed interim solution in this PR?

pangratz:
I haven't followed the latest developments in ember-cli, but have there been any changes regarding the addon-dependency-situation?

ffaubry:
Very interested by this PR... what's the status?

Cryrivers:
same question as above.

lucasm-iRonin:
What's the status? I would like to integrate [babel-plugin-typecheck](https://github.com/codemix/babel-plugin-typecheck) to my Ember apps and allowing for custom babel plugins is the only option I'm aware of.

Iftahh:
+1  
very interested in this

tschoartschi:
I'm also interested in the status of this pull request :)

mguida22:
An update on this would be great! 😄

tschoartschi:
I would really need an update on this issue. Because if it is not going to happen in the near future I'll have to rearrange my work packages. It is no problem but I just need to know. Thanks :)

tschoartschi:
@mguida22 @lucasm-iRonin @Cryrivers @ffaubry @Iftahh what I ended up doing is the following: I created an in-repo-addon and applied my babel-transformations after the ember-cli-babel plugin and before ember-cli-uglify. For me this works fine, maybe this is an solution for you as well

rwjblue:
Going to close this for now...

PR #48, opened 2015-07-25
explicitly filter the file we actually care about.

stefanpenner:


dfreeman:
:+1: Thanks for fixing this; sorry you had to!

PR #55, opened 2015-08-25
Fix ref to persistent filter

SlexAxton:
This is throwing weird git ref stuff on installs now. Likely due to the misspelling. For folks who might be locked in, it could be good to actually create a misspelled branch, so it just starts working....


SlexAxton:
Oh man, I got scooped like 3 times before this merged. Apologies. Still might be a good idea to create that branch? idk.

https://github.com/babel/ember-cli-babel/pull/51/files

PR #63, opened 2015-11-12
includePolyfill -> browserPolyfill

### LINKED ISSUE #61: includePolyfill option
I just upgraded ember-cli from 1.13.8 to master and started to get error:

```
Build failed.
File: ember-cli-app-version/components/app-version.js
[BABEL] ember-cli-app-version/components/app-version.js: Unknown option: direct.includePolyfill
```

It goes away if I remove `includePolyfill: true` from ember-cli-build.js, but I need the polyfill. It looks like the addon removes `options.includePolyfill` so not sure at all why this is an issue.

Dhaulagiri:
Fixes #61

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -5,7 +5,7 @@ var EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 var app = new EmberAddon({
   babel: {
-    includePolyfill: true
+    browserPolyfill: true
   }
 });
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ and [core.js](https://github.com/zloirock/core-js). Many transformations will wo
 must include the polyfill in your app. The [Babel feature tour](https://babeljs.io/docs/tour/) includes a note for
 features that require the polyfill to work.
 
-To include it in your app, pass `includePolyfill: true` in your `babel` options.
+To include it in your app, pass `browserPolyfill: true` in your `babel` options.
 
 ### About Modules
 
--- a/index.js
+++ b/index.js
@@ -25,7 +25,7 @@ module.exports = {
 
   shouldIncludePolyfill: function() {
     var options = getAddonOptions(this);
-    return options.includePolyfill === true;
+    return options.browserPolyfill === true;
   },
 
   importPolyfill: function(app) {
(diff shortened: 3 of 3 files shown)


rwjblue:
This seems to just be changing the flag that is supposed to be used, but the functionality seems to be the same.

ivanvotti:
@rwjblue I can confirm that this really fixes the bug #61, when you use `ember-cli-babel` with `ember-cli` 1.13.11 and above. I'm not sure if it's a good way to fix it, but it just works.

`ember-cli-babel` depends on `broccoli-babel-transpiler` that uses `browserPolyfill` as the option name.

Dhaulagiri:
@rwjblue I will try to circle back tomorrow to give better context around why this makes a difference if no one else beats me to it.

rwjblue:
> I can confirm that this really fixes the bug #61, when you use ember-cli-babel with ember-cli 1.13.11 and above.

I have no doubts that it works.  Neither this library nor broccoli-babel-transpiler have had a major version bump, and changing an option flag within the same major version seems to be a breaking change that we would want to avoid.  I just want to understand why this fixes the problem so that we can ensure that this is not a breaking change.

>  I will try to circle back tomorrow to give better context around why this makes a difference if no one else beats me to it.

@Dhaulagiri - Awesome, thank you!

Dhaulagiri:
I wonder if the real issue here is that previous release of ember-cli [did not use](https://github.com/ember-cli/ember-cli/commit/ee48665114a2d2dd71d785ef7689e90c9c5fda5a) broccoli-babel-transpiler.  Now that it does, that library is not happy that `includePolyfill` is coming through to its `options` hash, likely from [this line](https://github.com/ember-cli/ember-cli/blob/master/lib/broccoli/ember-app.js#L1022) in ember-cli.

ember-cli-babel makes an attempt to remove `includePolyfill` from the `options` hash [here](https://github.com/babel/ember-cli-babel/blob/master/index.js#L79) but it's not clear to me that this is actually doing anything since it is working with a cloned copy of the addon's options.

pixelhandler:
This fixes my build issue with ember-cli v1.13.12

Dhaulagiri:
This PR is a nice workaround but it doesn't seem to be the appropriate solution.  I'm not really sure what is though...

Dhaulagiri:
Closing in favor of the fix in https://github.com/ember-cli/ember-cli/pull/5104

jrjohnson:
@Dhaulagiri - can you keep your branch around for a few days until a new ember-cli is released?  It's a nice workaround to just pull from it in package.json

Dhaulagiri:
@jrjohnson we are using it too so I will keep it up for some time just to give people time to move off if after 1.13.13 is released

PR #65, opened 2015-11-19
Update README to have ES6 feature set

patocallaghan:
I needed to communicate to my team what ES6 features they could use without polyfills enabled and I put together the following list from the [Babel Docs](http://babeljs.io/docs/learn-es2015/)

@rwjblue suggested I add it to the README here


rwjblue:
I definitely like.  We will need to update when we transition to Babel 6, but that seems fine to me.

@stefanpenner - Any objections?

PR #67, opened 2015-12-13
Update options link in README to link to Babel 5.x options

BrianSipple:
Babel's update to 6.x brought with it significant changes to the  [options](https://babeljs.io/docs/usage/options/) passed during the transform. The `ember-cli-babel` `options` hash is still wired for Babel's 5.x [options](https://github.com/babel/babel.github.io/blob/5.0.0/docs/usage/options.md), whose docs, as it turns out, now take some considerable digging to find.

...So I thought I'd update the README to point in the right location once I uncovered it.


rwjblue:
:+1:

stefanpenner:
thanks, ya I plan to work on babel 6 support soon, just lots of user-land babel plugins need an overhaul to work. So that has been put on the back burner for now.

PR #70, opened 2016-02-29
Fix app options

artursmirnov:
When you define a `babel` section in `ember-cli-build.js` in order to pass options to babel transpiller, the addon does not find it because `addonContext.app` is not really an `EmberApp` instance. 
`EmberApp` instance is stored at `addonContext.app.app` and it has all the options.

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,7 +60,10 @@ module.exports = {
 };
 
 function getAddonOptions(addonContext) {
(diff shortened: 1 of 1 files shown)


artursmirnov:
I'm not sure why it doesn't run tests. The change should not brake anything.

artursmirnov:
Closing the PR for further investigation

PR #72, opened 2016-04-20
Add fastboot check

kolybasov:
This PR disabling `app.import` that is not allowed in ember fastboot. Maybe there are another way to fix this then I am ready to implement it, just let me know.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,8 @@ module.exports = {
 
   shouldIncludePolyfill: function() {
     var options = getAddonOptions(this);
-    return options.includePolyfill === true;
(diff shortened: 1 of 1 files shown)


stefanpenner:
Im not sure this is how we want to coordinate these two concepts, hardcoding fastboot in this addon seems funky.

@tomdale any thoughts?

tomdale:
@stefanpenner Not sure of a better option at the moment; ideally we will eventually have a first-calss notion of "targets" in Ember CLI. I think that is the right approach long-term.

kolybasov:
Not relevant anymore

PR #73, opened 2016-04-28
Fix link to babel docs

vitch:


Turbo87:
@vitch thanks! and sorry for the long response time...

PR #74, opened 2016-05-04
Fix babel docs link

elwayman02:


Turbo87:
looks like this is a duplicate of  #73

PR #83, opened 2016-08-05
Adds .npmignore and whitelists js files

twokul:
Makes sure to ignore files that are irrelevant when installing the package.

related to https://github.com/ember-cli/ember-cli/issues/6147


package.json
```diff
@@ -11,6 +11,10 @@
     "start": "ember server",
     "test": "ember test"
   },
+  "files": [
+    "index.js",
+    "ember-cli-build.js"
```
stefanpenner: I don't believe this file is required

package.json
```diff
@@ -11,6 +11,10 @@
     "start": "ember server",
     "test": "ember test"
   },
+  "files": [
+    "index.js",
```
stefanpenner: This one is

PR #86, opened 2016-08-11
Pass console object in to broccoli-babel-transpiler.

rwjblue:
This will be used by broccoli-babel-transpiler to log warnings/deprecations to the console when a given plugin does not provide the correct caching information.

This will be used by https://github.com/babel/broccoli-babel-transpiler/pull/89.

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,6 +69,21 @@ function getAddonOptions(addonContext) {
 function getBabelOptions(addonContext) {
   var options = clone(getAddonOptions(addonContext));
 
+  // pass a console object that wraps the addon's `UI` object
+  options.console = {
+    log: function(message) {
+      addonContext.ui.writeInfoLine(message);
+    },
+
+    warn: function(message) {
+      addonContext.ui.writeWarnLine(message);
+    },
+
+    error: function(message) {
+      addonContext.ui.writeError(message, 'ERROR');
+    }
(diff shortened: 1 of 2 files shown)


stefanpenner:
sg, in future versions of CLI we may want to have `ui.toConsole()` or something

rwjblue:
Ya, that seems like a good idea.

stefanpenner:
for the future.

rwjblue:
Build is failing because currently there is no way to distinguish options intended to be passed into `babel` itself, versus those intended for `broccoli-babel-transpiler` and when `babel` gets the `console` object passed here it throws an error.

Ideally, we should refactor the babel options to be passed in via `options.babel` or something, so that the other non-babel options (like `console`, `includePolyfill`, etc) can be in the root of `options`.

rwjblue:
I guess I have to release broccoli-babel-transpiler first, then this...

rwjblue:
Released broccoli-babel-transpiler@5.6.0 and rebased this to update to that version.

rwjblue:
@stefanpenner / @Turbo87 - This is ready to go, but I don't have merge + release access here.  If one of y'all could release (or add me) I would appreciate it...

stefanpenner:
releasing as v5.1.7

PR #104, opened 2016-12-05
Replace deprecated version checker method call

Turbo87:
same pattern as in https://github.com/ember-cli/ember-cli-mocha/blob/v0.13.0/index.js#L69-L72

/cc @rwjblue @stefanpenner


rwjblue:
LGTM

PR #105, opened 2016-12-05
Move custom options to "ember-cli-babel" options hash

### LINKED ISSUE #102: Deprecate custom properties in `babel` options hash
As mentioned in https://github.com/ember-cli/ember-cli/issues/5103 we are currently reading custom properties (`includePolyfill` and `compileModules`) from the `babel` options hash, then deleting them from the hash, and finally passing that hash to `broccoli-babel-transpiler`. Instead we should put those options in a `ember-cli-babel` options hash, and pass the `babel` hash directly to the transpiler.

This should be fixed in a backwards compatible way by deprecating the use of those custom properties in the `babel` hash and making **both** work until the next major version bump.

Since `ember-cli` is currently [setting the `compileModules` option](https://github.com/ember-cli/ember-cli/blob/v2.10.0/lib/broccoli/ember-app.js#L192) and we don't want to flood our users with deprecation message that they can't fix we should only show those warnings if `ember-cli > 2.10`. We should also set some sort of `customConfigKey: 'ember-cli-babel'` flag so that `ember-cli` knows what config key to use.

/cc @rwjblue

Turbo87:
Resolves #102 and unblocks part of the journey to Babel 6

/cc @rwjblue @stefanpenner @nathanhammond

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,16 +37,17 @@ module.exports = {
     var babelOptions = addonOptions.babel;
     var customOptions = addonOptions['ember-cli-babel'];
 
+    if (this._shouldShowBabelDeprecations && !this._polyfillDeprecationPrinted &&
+      babelOptions && 'includePolyfill' in babelOptions) {
+
+      this._polyfillDeprecationPrinted = true;
+      this.ui.writeDeprecateLine(
+        'Putting the "includePolyfill" option in "babel" is deprecated, please put it in "ember-cli-babel" instead.');
+    }
+
     if (customOptions && 'includePolyfill' in customOptions) {
       return customOptions.includePolyfill === true;
-
     } else if (babelOptions && 'includePolyfill' in babelOptions) {
-      if (this._shouldShowBabelDeprecations && !this._polyfillDeprecationPrinted) {
-        this._polyfillDeprecationPrinted = true;
-        this.ui.writeDeprecateLine(
-          'Putting the "includePolyfill" option in "babel" is deprecated, please put it in "ember-cli-babel" instead.');
-      }
-
       return babelOptions.includePolyfill === true;
     } else {
       return false;
@@ -97,16 +98,17 @@ module.exports = {
     var babelOptions = addonOptions.babel;
     var customOptions = addonOptions['ember-cli-babel'];
 
+    if (this._shouldShowBabelDeprecations && !this._modulesDeprecationPrinted &&
+      babelOptions && 'compileModules' in babelOptions) {
+
+      this._modulesDeprecationPrinted = true;
+      this.ui.writeDeprecateLine(
+        'Putting the "compileModules" option in "babel" is deprecated, please put it in "ember-cli-babel" instead.');
+    }
+
     if (customOptions && 'compileModules' in customOptions) {
       return customOptions.compileModules === true;
-
     } else if (babelOptions && 'compileModules' in babelOptions) {
-      if (this._shouldShowBabelDeprecations && !this._modulesDeprecationPrinted) {
-        this._modulesDeprecationPrinted = true;
-        this.ui.writeDeprecateLine(
-          'Putting the "compileModules" option in "babel" is deprecated, please put it in "ember-cli-babel" instead.');
-      }
-
       return babelOptions.compileModules === true;
     } else {
       return false;
--- a/index.js
+++ b/index.js
@@ -39,6 +39,7 @@ module.exports = {
 
     if (customOptions && 'includePolyfill' in customOptions) {
       return customOptions.includePolyfill === true;
+
     } else if (babelOptions && 'includePolyfill' in babelOptions) {
       if (this._shouldShowBabelDeprecations && !this._polyfillDeprecationPrinted) {
         this._polyfillDeprecationPrinted = true;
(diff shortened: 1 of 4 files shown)


nathanhammond:
- Tests? I think we can unit test the invocations to `_getBabelOptions` and `shouldIncludePolyfill`.
- Are there other options we supported for `ember-cli-babel` under the `babel` namespace?

Turbo87:
> Are there other options we supported for ember-cli-babel under the babel namespace?

since those are the only two options that `ember-cli-babel` cares about I don't think there are more

> Tests? I think we can unit test the invocations to _getBabelOptions and shouldIncludePolyfill

do we have some sort of test harness for that?

Turbo87:
@nathanhammond got tests now for the new logic

index.js
```diff
@@ -25,14 +27,30 @@ module.exports = {
...
 
   shouldIncludePolyfill: function() {
-    var options = getAddonOptions(this);
-    return options.includePolyfill === true;
+    var addonOptions = this._getAddonOptions();
+    var babelOptions = addonOptions.babel;
+    var customOptions = addonOptions['ember-cli-babel'];
+
+    if (customOptions && 'includePolyfill' in customOptions) {
+      return customOptions.includePolyfill === true;
```
nathanhammond: Can we check to see if it is in both places here as well and `writeDeprecateLine`?

index.js
```diff
@@ -69,74 +87,96 @@ module.exports = {
...
-    },
-
-    error: function(message) {
-      // fallback needed for support of ember-cli < 2.2.0
-      if (ui.writeError) {
-        ui.writeError(message);
-      } else {
-        ui.writeLine(message, 'ERROR');
+  _getAddonOptions: function() {
+    return (this.parent && this.parent.options) || (this.app && this.app.options) || {};
```
nathanhammond: That second check shouldn't be required unless we're going _way_ back in support. (Before 0.2.0.)

index.js
```diff
@@ -69,74 +87,96 @@ module.exports = {
...
+  },
+
+  _shouldCompileModules: function(addonOptions) {
+    var babelOptions = addonOptions.babel;
+    var customOptions = addonOptions['ember-cli-babel'];
+
+    if (customOptions && 'compileModules' in customOptions) {
+      return customOptions.compileModules === true;
```
nathanhammond: Same story: can we check to see if it is in both places here as well and `writeDeprecateLine`?

index.js
```diff
@@ -69,74 +87,96 @@ module.exports = {
...
+    }
+  },
 
-  // do not enable non-standard transforms
-  if (!('nonStandard' in options)) {
-    options.nonStandard = false;
-  }
+  _getBabelOptions: function() {
+    var addonOptions = this._getAddonOptions();
+    var options = clone(addonOptions.babel || {});
```
nathanhammond: Can we name `options` a bit more clearly? I had to look at this multiple times to figure out what it was supposed to be.

index.js
```diff
@@ -69,74 +87,96 @@ module.exports = {
...
 
+    // Remove custom options from `options` hash that is passed to Babel
+    delete options.includePolyfill;
     delete options.compileModules;
-  } else {
-    if (options.blacklist.indexOf('es6.modules') < 0) {
-      options.blacklist.push('es6.modules');
+
+    if (compileModules) {
+      if (options.blacklist.indexOf('es6.modules') >= 0) {
```
nathanhammond: Let's save this off, you use it three times.

nathanhammond:
After these changes I'm pretty sure it's good to go. Would like a bonus confirmation from @rwjblue.

index.js
rwjblue: FYI - we can't rely on writeDeprecateLine as far back as this lib supports

index.js
```diff
@@ -16,6 +17,7 @@ module.exports = {
     var dep = checker.for('ember-cli', 'npm');
 
     this._shouldSetupRegistryInIncluded = !dep.satisfies('>=0.2.0');
+    this._shouldShowBabelDeprecations = !dep.lt('2.11.0-beta.1');
```
rwjblue: I think this should be `< 2.11.0-beta.2` since 2.11.0-beta.1 didn't include the code to prevent deprecations...?

rwjblue:
Agreed, I am 👍 once the current round of requests are resolved.  We should followup with updates in ember-cli's beta branch ASAP though...

index.js
Turbo87: @nathanhammond that is pretty much the original code and not modified by me (see https://github.com/babel/ember-cli-babel/blob/b9a0dc232f4038bf6c95de51f71b04faff08406d/index.js#L76)

index.js
Turbo87: `options` was the original name. I kept it to try to keep the diff smaller.

index.js
Turbo87: again: this existed already before this PR (see https://github.com/babel/ember-cli-babel/blob/b9a0dc232f4038bf6c95de51f71b04faff08406d/index.js#L125-L135)

index.js
Turbo87: @rwjblue I'm fine with either

index.js
Turbo87: @rwjblue since the deprecation will only be triggered in recent CLIs we can just use `writeDeprecateLine` here

@nathanhammond I'll move the deprecation check in front of everything

Turbo87:
@rwjblue @nathanhammond addressed your requests

index.js
rwjblue: Good point RE, newer CLI's only doing the deprecation. Maybe you can leave an inline comment (so future selves don't see it's use here as an excuse to use it for other ember-CLI deprecations)? I don't feel strongly either way though...

Turbo87: done

index.js
nathanhammond: 👍

index.js
nathanhammond: Yes, but you already modify those lines, so let's address this one.

Turbo87: done

PR #106, opened 2016-12-07
Fix feature detection bug for setupPreprocessorRegistry.

rwjblue:
The logic here was incorrect for beta versions of ember-cli (due to bizarreness in how `semver` package deals with prereleases).

This fixes the issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@ module.exports = {
     var checker = new VersionChecker(this);
     var dep = checker.for('ember-cli', 'npm');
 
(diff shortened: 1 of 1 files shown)


Turbo87:
@rwjblue the issue here was actually that I moved from the deprecated version checker functions to using `checker.for('ember-cli', 'npm')`, which behaves differently. To be precise the NPM-based checker includes the prerelease stuff, while the deprecated CLI checker doesn't (see https://github.com/ember-cli/ember-cli-version-checker/blob/master/src/ember-cli-dependency-version-checker.js#L12-L13)

PR #111, opened 2017-02-08
Fix for undefined parent error

hsdwayne:
This fixes the error discussed in this comment https://github.com/babel/ember-cli-babel/commit/74c47232677d6153226187dbb1648cfa0dcb9250#commitcomment-20790735

### DIFF
--- a/index.js
+++ b/index.js
@@ -120,14 +120,23 @@ module.exports = {
   },
 
   _getBabelOptions: function() {
-    var parentName = typeof this.parent !== 'undefined' ? (typeof this.parent.name === 'function' ? this.parent.name() : this.parent.name) : 'unknown';
+    var parentName;
+
+    if (this.parent) {
+      if (typeof this.parent.name === 'function') {
+        parentName = this.parent.name();
+      } else {
+        parentName = this.parent.name;
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -120,7 +120,7 @@ module.exports = {
   },
 
   _getBabelOptions: function() {
-    var parentName = typeof this.parent.name === "function" ? this.parent.name() : this.parent.name;
+    var parentName = typeof this.parent !== 'undefined' ? (typeof this.parent.name === 'function' ? this.parent.name() : this.parent.name) : 'unknown';
```
rwjblue: Maybe a slight tweak:

```js
var parantName;
if (this.parent && typeof this.parent.name === 'function') {
  parentName = this.parent.name();
} else if (this.parent) {
  parentName = this.parent.name;
}
```

I find it hard to reason about nested ternaries...

index.js
```diff
@@ -120,14 +120,23 @@ module.exports = {
...
+    var parentName;
+
+    if (this.parent) {
+      if (typeof this.parent.name === 'function') {
+        parentName = this.parent.name();
+      } else {
+        parentName = this.parent.name;
+      }
+    }
+
```
hsdwayne: @rwjblue Ok, how's this?

PR #113, opened 2017-02-21
Update README.md

marpo60:
* `includePolyfill` is no longer part of `babel` options, now is part of `ember-cli-babel` options.
* A deprecation is shown if it was used in `babel` options
  https://github.com/babel/ember-cli-babel/blob/2b1c8cf06937d0f37bfb378cbf435c721848f837/index.js#L47


rwjblue:
👍

rwjblue:
Thank you!

PR #117, opened 2017-03-13
Fix issues with isPluginRequired.

rwjblue:


cibernox:
Having ember-cli as a middleman to babel-preset-env is fantastic. 
Should we document this feature or should we bikeshed the API a bit more?

rwjblue:
I need to do a thorough revamp of the README for babel@6 stuff, this should be included...

PR #122, opened 2017-03-21
Properly forward the browsers targets to ember-preset-env

kanongil:
Currently they are completely ignored.


cibernox:
I just wonder if we can create a test harness that ember-cli-babel honour targets.

rwjblue:
@kanongil - How embarrassing, thank you for digging and submitting the fix!

@cibernox - Ya, we have a bunch of node-tests already, I'll add one to confirm preset env is generally working (pass a specific set of targets and confirm various plugins are not used).

PR #124, opened 2017-03-21
Add basic sanity test to confirm babel-preset-env is working.

rwjblue:
Confirms that a specific plugin is or is not used based on the specified targets.


rwjblue:
@cibernox - FYI, this adds a test that sanity checks that preset env is actually working.  It fails before the changes made in #122.

PR #144, opened 2017-05-15
upgrade all the deps

stefanpenner:


.travis.yml
```diff
@@ -1,22 +1,37 @@
...
-  - npm install -g bower
-  - npm install
-  - bower install
+  - yarn install --no-lockfile
 
 script:
-  - npm test
+  # Usually, it's ok to finish the test scenario without reverting
+  #  to the addon's original dependency state, skipping "cleanup".
+  - node_modules/.bin/ember try:one $EMBER_TRY_SCENARIO test --skip-cleanup
```
rwjblue: I believe this will result in us not actually running the node tests?

.travis.yml
```diff
@@ -1,22 +1,37 @@
...
-  - "7"
+
+sudo: false
+
+cache:
+  yarn: true
+
+env:
+  # we recommend testing LTS's and latest stable release (bonus points to beta/canary)
+  - EMBER_TRY_SCENARIO=ember-lts-2.4
```
rwjblue: There is no ember specific code here, I don't think we need to test more than one (slowing down CI runs...)

.travis.yml
```diff
@@ -1,22 +1,37 @@
 ---
-sudo: false
-
 language: node_js
 node_js:
-  - "4"
   - "6"
-  - "7"
```
rwjblue: We need to continue to test against 4, 6, and 7.

stefanpenner: Oops

stefanpenner:
@rwjblue thanks for the review, should be addressed now!

Issue #147, opened 2017-05-31
NodeJS 8.0.


johnunclesam:
```
error ember-cli-babel@6.1.0: The engine "node" is incompatible with this module. Expected version "^4.5 || 6.* || 7.*".
error Found incompatible module
```

stefanpenner:
updating CI now

stefanpenner:
fixed on master actually: https://github.com/babel/ember-cli-babel/blob/master/package.json#L72

Likley just needs a release

johnunclesam:
yes, please.

stefanpenner:
https://github.com/babel/ember-cli-babel/pull/148

stefanpenner:
merged, but we are having some issues releasing to NPM: https://github.com/babel/ember-cli-babel/issues/153

PR #148, opened 2017-05-31
[BUGFIX #147] add node 8 support

stefanpenner:


node-tests/addon-test.js
```diff
@@ -691,7 +691,7 @@ describe('ember-cli-babel', function() {
       expect(pluginRequired).to.be.true;
     });
 
-    it('returns false when targets do not require plugin', function() {
+    it.only('returns false when targets do not require plugin', function() {
```
existentialism: @stefanpenner sorry about that... was (foolishly) not even considering anyone else using `isPluginRequired` when we made string version target changes (originally targeted for 2.x) :(

stefanpenner: @existentialism not a problem. Just a surprise, glad it was easy to work around.

PR #152, opened 2017-05-31
[FIXES #150] update babel-preset-env

### LINKED ISSUE #150: update babel-preset-env
They made a breaking change, so I locked us to 1.3.3 for now https://github.com/babel/ember-cli-babel/pull/148/commits/5b52b660fc1e8d15af8913a42ed13e16c6cbf8d1

We should bump, but it will require slight adjustments to: https://github.com/babel/ember-cli-babel/blob/master/index.js#L248 some of which has been already done [here](https://github.com/babel/ember-cli-babel/pull/140)

stefanpenner:


index.js
```diff
@@ -225,7 +225,7 @@ module.exports = {
     let browsers = targets && targets.browsers;
     let presetOptions = Object.assign({}, options, {
       modules: false,
-      targets: { browsers },
+      targets: targets
```
rwjblue: Can probably just use shorthand since the names are the same

stefanpenner: does that work in node 4?

stefanpenner: yup

PR #158, opened 2017-06-19
Respect babel sourceMaps option

dwickern:
The `babel.sourceMaps` option stopped working since the switch to ember-cli-babel. The option is used to enable ES6 sources in the debugger.

This PR enables sourceMaps under the `ember-cli-babel.sourceMaps` config option and also `babel.sourceMaps` for backwards compatibility.

Originally implemented in ember-cli/ember-cli#6458
Fixes (again) ember-cli/ember-cli#6066

### DIFF
--- a/index.js
+++ b/index.js
@@ -192,8 +192,14 @@ module.exports = {
 
     let providedAnnotation = config['ember-cli-babel'] && config['ember-cli-babel'].annotation;
 
+    let sourceMaps = false;
+    if (config.babel && 'sourceMaps' in config.babel) {
+      sourceMaps = config.babel.sourceMaps;
+    }
+
     let options = {
-      annotation: providedAnnotation || `Babel: ${this._parentName()}`
+      annotation: providedAnnotation || `Babel: ${this._parentName()}`,
+      sourceMaps
     };
 
     let userPlugins = addonProvidedConfig.plugins;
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -190,10 +190,12 @@ module.exports = {
...
     let shouldCompileModules = this._shouldCompileModules(config);
 
-    let providedAnnotation = config['ember-cli-babel'] && config['ember-cli-babel'].annotation;
+    let addonOptions = config['ember-cli-babel'] || {};
+    let babelOptions = config.babel || {};
 
     let options = {
-      annotation: providedAnnotation || `Babel: ${this._parentName()}`
+      annotation: addonOptions.annotation || `Babel: ${this._parentName()}`,
+      sourceMaps: addonOptions.sourceMaps || babelOptions.sourceMaps
```
rwjblue: Hmm, I'm not sure that we should do this sort of fallback. Each option needs to have a single place to live, if we have to support both locations one should be deprecated.

dwickern: I thought `babel` *was* deprecated. At least, I remember fixing a deprecation error when I upgraded to ember-cli 2.12

rwjblue: Both are still valid, but each of them for different types of options.

At this point, I would say that `sourceMaps` should go in `babel` config (since we do pass it through directly to babel itself), and that we shouldn't allow it to be specified in `options['ember-cli-babel']`.

dwickern: Sounds good. I won't bother with a warning or deprecation message since `sourceMaps` was never valid inside `ember-cli-babel` to begin with

rwjblue: Perfect!

thefonso:
can someone point me to instructions on how to use this?

dwickern:
@thefonso https://github.com/babel/ember-cli-babel#enabling-source-maps

PR #159, opened 2017-07-03
Add emberjs/rfcs#176 support by default.

rwjblue:
This replaces the new modules with globals usage when the Ember version in use does not include its own modules.

As discussed in a recent ember-cli team meeting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ interface EmberCLIBabelConfig {
     includePolyfill?: boolean;
     compileModules?: boolean;
     disableDebugTooling?: boolean;
+    disableEmberModulesAPIPolyfill?: boolean;
   };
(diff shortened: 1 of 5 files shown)


Turbo87:
@rwjblue I suppose we won't backport this to Babel 5, right?

rwjblue:
@Turbo87 - I wasn't planning to, but we _could_ if you thought it was required...

Turbo87:
likely not necessary since we can mix addons with 5 and 6 in an app and this will create another incentive to update to Babel 6

rwjblue:
Sounds good.  We also did a bunch of work to make ember-cli-babel@6 work quite a ways back, so folks should not be stuck...

Issue #160, opened 2017-07-03
Uncaught ReferenceError: _asyncToGenerator is not defined


Turbo87:
I'm seeing the following error when building my app with `ember-cli-babel@6.5.0`:

```
Uncaught ReferenceError: _asyncToGenerator is not defined
```

The source file contains:

```
import { assert } from '@ember/debug';
```

and once I remove that line the `_asyncToGenerator()` is correctly included into the output.

PR #161, opened 2017-07-03
Avoid conflicting transforms for @ember/debug.

rwjblue:
When both babel-plugin-debug-macros and babel-plugin-ember-modules-api-polyfill attempt to replace `import { assert } from '@ember/assert';` the interaction between the two also removes any helpers that are injected by other parts of the transpilation (e.g. `_classCallCheck` or `_asyncToGenerator`) therefore making the transpiled output fail at runtime.

This updates babel-plugin-ember-modules-api-polyfill to a version that allows specifying a blacklist so that we can avoid this conflict.

Fixes https://github.com/babel/ember-cli-babel/issues/160

### DIFF
--- a/index.js
+++ b/index.js
@@ -260,7 +260,7 @@ module.exports = {
     if (this._emberVersionRequiresModulesAPIPolyfill()) {
       const ModulesAPIPolyfill = require('babel-plugin-ember-modules-api-polyfill');
 
-      return [[ModulesAPIPolyfill, { blacklist: ['@ember/debug'] }]];
+      return [[ModulesAPIPolyfill, { blacklist: { '@ember/debug': ['assert', 'deprecate', 'warn']} }]];
     }
   },
 
--- a/index.js
+++ b/index.js
@@ -260,7 +260,7 @@ module.exports = {
     if (this._emberVersionRequiresModulesAPIPolyfill()) {
       const ModulesAPIPolyfill = require('babel-plugin-ember-modules-api-polyfill');
 
-      return [[ModulesAPIPolyfill]];
+      return [[ModulesAPIPolyfill, { blacklist: ['@ember/debug'] }]];
     }
   },
 
--- a/node-tests/addon-test.js
+++ b/node-tests/addon-test.js
@@ -144,6 +144,7 @@ describe('ember-cli-babel', function() {
 
         let contents = output.read()['foo.js'];
 
+        expect(contents).to.not.include('@ember/debug');
         expect(contents).to.include('function _asyncToGenerator');
         expect(contents).to.include('var inspect = Ember.inspect;');
         expect(contents).to.not.include('assert');
--- a/node-tests/addon-test.js
+++ b/node-tests/addon-test.js
@@ -96,7 +96,8 @@ describe('ember-cli-babel', function() {
 
       it("should replace imports by default", co.wrap(function* () {
         input.write({
-          "foo.js": `import Component from '@ember/component';`
+          "foo.js": `import Component from '@ember/component';`,
+          "app.js": `import Application from '@ember/application';`
         });
 
         subject = this.addon.transpileTree(input.path());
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -260,7 +260,7 @@ module.exports = {
     if (this._emberVersionRequiresModulesAPIPolyfill()) {
       const ModulesAPIPolyfill = require('babel-plugin-ember-modules-api-polyfill');
 
-      return [[ModulesAPIPolyfill]];
+      return [[ModulesAPIPolyfill, { blacklist: ['@ember/debug'] }]];
```
Turbo87: are https://github.com/ember-cli/ember-rfc176-data/blob/master/globals.json#L74-L80 all supported in the other Babel plugin?

rwjblue: I don't think all of those are required (e.g. this blacklist doesn't prevent `@ember/debug/data-adapter` or `@ember/debug/container-debug-adapter`).  The only one that is not supported by babel-plugin-debug-macros is `Ember.inspect` (which seems misplaced in `@ember/debug` IMHO).

Turbo87: so `import { inspect } from '@ember/debug';` won't work at all after this PR?

rwjblue:
Updated to properly allow named exports from `@ember/debug` to be handled by the "proper" plugin (and added a test to confirm). 

@Turbo87 - Ready for another look...

stefanpenner:
oh babel...

Turbo87:
It looks like `@ember/debug` is still not transpiled properly with the update:

```
define('ct/controllers/search', ['exports', 'ember', 'ct/utils/pnr_utils', 'ct/computed', 'ct/utils/transition', 'ct/utils/book_context', 'ct/constants', 'ct/mixins/errors_support', '@ember/debug'], function (exports, _ember, _pnr_utils, _computed, _transition, _book_context, _constants, _errors_support) {
```

From what I understand I shouldn't see an `@ember/debug` import here at all, but even if I should then there should be a corresponding parameter in the function definition, right?

Turbo87:
I *think* we might have to change the order of:

```js
       this._getDebugMacroPlugins(config),
       this._getEmberModulesAPIPolyfill(config),
```

rwjblue:
Ordering is totally screwed up. Changing the order actually has no effect due to the way babel merges all plugins into a single AST visitor.

rwjblue:
@turbo87 - Can you share the raw source of the imports for that snippet? Can you also confirm if the thing imported from `@ember/debug` is actually being used?

rwjblue:
@Turbo87 - OK, fixed that last issue I believe. Can you give it a try again?

For more info on what was wrong, you can see the writeup in https://github.com/chadhietala/babel-plugin-debug-macros/issues/45.

Turbo87:
@rwjblue sorry to disappoint you... 😞 

it now correctly transpiled the debug macros things, but it no longer converts the module imports at all 🤔 

```
define('ct/app', ['exports', '@ember/application', 'ember', 'ct/resolver',
  'ember-load-initializers', 'ct/environment'], function (exports, _application,
  _ember, _resolver, _emberLoadInitializers, _environment) {
 ```

rwjblue:
@Turbo87 - I added a few more tests/assertions around non-`@ember/debug` usage just to add a bit more coverage. I'm trying to reproduce, but can't seem to...

Turbo87:
looks like clearing my caches resolved the issue. this seems good to go 👍

Dhaulagiri:
(pardon my rudimentary understanding) but with this change if try an import like `import { warn } from '@ember/debug';` in an addon I'm working with running ember-cli-babel 6.8.2 , `warn` is undefined because its transpilation back to an Ember global is blacklisted.

If it's helpful I can open a new issue and try to make a reproduction, but perhaps there is something I'm not understanding about the intent of this PR?

rwjblue:
@Dhaulagiri - We have two plugins that both know how to operate on `import { warn } from '@ember/debug';`. The first is `babel-plugin-ember-modules-api-polyfill` which will remove the import and replace (essentially) with `Ember.*` usage.  The second is `babel-plugin-debug-macros` which will replace these specific debug statements with customized debug build only output. This PR is all about preventing these two plugins from conflicting by disabling the modules-api-polyfill (because we want that prod build stripping yumminess).

rwjblue:
Checkout [this test](https://github.com/babel/ember-cli-babel/blob/d02675c42b28ad571f196748a3e1ed28c0ca1c36/node-tests/addon-test.js#L211-L231) for example output (and confirmation that we are indeed processing these).

PR #164, opened 2017-07-09
Improve acceptance tests and Ember syntax

villander:


tests/dummy/app/router.js
```diff
@@ -1,12 +1,14 @@
 import Ember from 'ember';
 import config from './config/environment';
 
-const Router = Ember.Router.extend({
+const { Router } = Ember;
+
+const EmRouter = Router.extend({
```
kellyselden: I would prefer to stay in sync with the official blueprint.

tests/dummy/app/controllers/application.js
```diff
@@ -1,9 +1,12 @@
 import Ember from 'ember';
 
-export default Ember.Controller.extend({
+const { Controller, A, computed } = Ember;
```
kellyselden: This is going to be outdated syntax extremely soon now that addons are starting to update to the new module api.

tests/acceptance/simple-test.js
```diff
@@ -10,16 +12,16 @@ module('Simple Acceptance Test', {
...
-test('value of input', function(assert) {
+test('ES6 features work correcly', function(assert) {
   visit('/');
 
   andThen(() => {
-    assert.equal('Test Value', find('#test-input').val());
-    assert.equal('one', find('#first-value').text());
-    assert.equal('two', find('#last-value').text());
+    assert.equal('Test Value', find('#test-input').val(), 'Has arrow functions and template string as ES6 feature');
+    assert.equal('one', find('#first-value').text(), 'Has generatos as ES6 feature');
```
kellyselden: generatos => generators

tests/acceptance/simple-test.js
```diff
@@ -10,16 +12,16 @@ module('Simple Acceptance Test', {
   },
 
   afterEach() {
-    Ember.run(App, 'destroy');
+    run(App, 'destroy');
   }
 });
 
-test('value of input', function(assert) {
+test('ES6 features work correcly', function(assert) {
```
kellyselden: all these text clarifications I like.

tests/dummy/app/router.js
villander: @kellyselden But the official blueprint has a global sweetening, why not follow the Dockyard styleguide where we reduce global calls?

https://github.com/DockYard/styleguides/blob/master/engineering/ember.md#import-what-you-use-do-not-use-globals

tests/dummy/app/controllers/application.js
villander: Ok I'll go back to the original

tests/dummy/app/router.js
kellyselden: That styleguide will soon be out of date. This is the new importing standard: https://github.com/ember-cli/ember-rfc176-data

villander: @kellyselden Can I make that change then? From which ember-cli is it already available? I've seen that ember-basic-dropdown is already using it, although it's still in beta.

https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/components/basic-dropdown/content.js#L1

villander:
Done @kellyselden

villander:
@kellyselden we can do merge this?

kellyselden:
I'm trying to get #167 merged first so we can add some linting rules.

rwjblue:
#167 has landed

PR #165, opened 2017-07-10
Include examples for where to put options

dfreeman:
Stemming from a conversation with @turbo87 in Slack last week, this updates the README to include specific examples of how to configure Babel options in an app, an engine, or a regular addon.


rwjblue:
Seems good, thank you!

PR #176, opened 2017-08-11
Passes option to skip Ember.String processing

locks:


villander:
@locks what the test? use case

I want help you on this

rwjblue:
@locks - Looks good, I just released the polyfill with the required changes, can you rebase and update the minimum version?

locks:
@villander we need to check that we are actually detecting the `@ember/string` and passing the appropriate option to the babel plugin.

index.js
```diff
@@ -253,14 +253,19 @@ module.exports = {
   },
 
   _getEmberModulesAPIPolyfill(config) {
+    const blacklist = { '@ember/debug': ['assert', 'deprecate', 'warn'] };
```
rwjblue: Can you extract the logic here into a `_getEmberModulesAPIBlacklist()` method?

locks: Addressed, was that what you had in mind?

rwjblue: Ya, thank you

locks:
We are G.T.G.™

index.js
```diff
@@ -347,5 +348,27 @@ module.exports = {
...
+      ];
+    }
+
+    return blacklist;
+  },
+
+  _emberStringDependencyPresent() {
+    let checker = new VersionChecker(this.parent).for('@ember/string', 'npm');
+
+    return checker.isAbove('0.0.0');
```
Turbo87: why `.isAbove()`? I think there is an `.exists()` now 🤔

rwjblue: Ya, confirm. But only recently IIRC.

PR #186, opened 2017-11-03
Updates link to the Babel mono-repo

alvincrespo:
When reading through the options I visited the link but found that the documented options were moved to the Babel mono-repo. This PR makes the appropriate change to ensure developers can access the documentation directly.


rwjblue:
Awesome, thanks @alvincrespo!

PR #190, opened 2017-11-15
Properly detect host ember-cli version.

rwjblue:
When running in a linked environment, the standard `require.resovle` technique to determine the currently running ember-cli version does not work properly (because the linked addon's ember-cli dev dependency will be found instead of the _actual_ running ember-cli from the project).

The most common symptom of this issue is that modules are "double transpiled" (e.g. you have an AMD wrapped module _inside_ of another AMD wrapper).

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@
 const VersionChecker = require('ember-cli-version-checker');
 const clone = require('clone');
 const path = require('path');
+const semver = require('semver');
 
 let count = 0;
 
@@ -339,7 +340,7 @@ module.exports = {
 
       return babelOptions.compileModules;
     } else {
-      return this.emberCLIChecker.gt('2.12.0-alpha.1');
+      return semver.gt(this.project.emberCLIVersion, '2.12.0-alpha.1');
     }
   },
 
--- a/index.js
+++ b/index.js
@@ -351,10 +351,10 @@ module.exports = {
   },
 
   _getEmberModulesAPIBlacklist() {
-    const blacklist = { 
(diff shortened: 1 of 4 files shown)


Turbo87:
shouldn't we fix this in ember-cli-version-checker instead?

rwjblue:
The issue is that it depends. If you are trying to detect the currently running processes version of ember-cli you do the thing I’m doing here, but Ember-cli-version-checker is mostly just checking what is _installed_ within the specific root. In other words, checking the running processes version of ember-cli isn’t always correct...

Turbo87:
> In other words, checking the running processes version of ember-cli isn’t always correct...

please elaborate 🤔

rwjblue:
It is correct in this case because the running processes Ember-cli will be controlling any build stuffs, however you could easily be running ember-cli-version-checker from within a process started (for example) with the globally installed ember-cli. In that scenario you want to detect the installed version, not the running one....

rwjblue:
I’m also not sure that I like the idea of special casing specific packages in the version checker (yes I know we made a special method for Ember itself). 

IMHO, the majority of the time, what it does now is correct, and apecial casing sucks (generally speaking).

I’m happy to move this to the version checker if you feel very strongly. I just want to unblock folks currently being trolled by the bug I’m fixing.

rwjblue:
Discussed with @Turbo87 in slack, we are both 👍 on fixing the user facing issue here first and will circle around later to decide if `ember-cli-version-checker` should be updated/tweaked in some way...

PR #199, opened 2018-01-14
Remove redundant check for targets

astronomersiva:
Hi,

Spotted this while going through the source of ember-cli-babel. Looks like it was introduced while [adding tests for isPluginRequired](https://github.com/babel/ember-cli-babel/commit/f6fabd72633a5d8174d2a07be6a93469b8513042#diff-168726dbe96b3ce427e7fedce31bb0bcR192).

### DIFF
--- a/index.js
+++ b/index.js
@@ -299,7 +299,7 @@ module.exports = {
   },
 
   _getTargets() {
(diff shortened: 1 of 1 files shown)


rwjblue:
Whoops! Thank you for fixing!

PR #204, opened 2018-02-18
add a no-op optimization

### LINKED ISSUE #201: detect when no plugins are added and no-op
Discovered as a new optimization from https://github.com/ember-cli/ember-cli/pull/7501.

cc @rwjblue

kellyselden:
Closes #201

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,8 +49,14 @@ module.exports = {
     let description = `000${++count}`.slice(-3);
     let postDebugTree = this._debugTree(inputTree, `${description}:input`);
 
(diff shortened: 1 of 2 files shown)


stefanpenner:
@kellyselden I like this, but can you add a quick unit test for that hook, ensuring that this works as expected (and continues to into the future)

kellyselden:
@stefanpenner added a test

kellyselden:
@rwjblue That's essentially what my test is. What would you like changed?

stefanpenner:
released as v6.12.0 🎉

PR #205, opened 2018-03-03
Respect babel retainLines option.

dan-tailormed:
Hi,

While trying to set ember developing environment with vscode and vscode-debug-chrome extenstion I was facing some issues with setting breakpoints since they where offset in the source code.

I found that this setting allow for correct line mapping and by this solving this issue.
I think that others might find this useful for their development as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,7 @@ interface EmberCLIBabelConfig {
     exclude?: string[];
     useBuiltIns?: boolean;
     sourceMaps?: boolean | "inline" | "both";
+    retainLines?: boolean;
     plugins?: BabelPlugin[];
   };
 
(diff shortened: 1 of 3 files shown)


stefanpenner:
Any reason to not enable this all the time? "semi wacky code" may not be bad, as long as it isn't TOO BAD.

Areas of concern would be:

* build speed
* ??

cc @rwjblue

rwjblue:
What does the option actually do? Can you share some before / after code snippets (and likely sourcemaps)?

dan-tailormed:
Hey,
Please see the following:

## When using original current version of babel setting a breakpoint will not stop on the correct line

original-babel-set-bp

original-babel-stop-at-wrong-line

## After installing the modified version that support retainLines

activate-retainline-in-ember-cli-build

set-bp-with-retainline-activated

stop-in-correct-line

## Settings in launch.js file to support this

debug-for-chrome-launch-settings

@stefanpenner I do not notice any major build speed change

@rwjblue This addition to support retainLines is optional and can be controlled by the developer from ember-cli-build.js

If you will accept this PR I'm sure it can assist many developers and should be documented in the readme.

Thanks

arthur5005:
Wondering if there's any movement on this, running into line offset problems often while developing.

PR #213, opened 2018-03-26
deprecationMessage tweaks

kellyselden:


goldenice:
Err, this breaks it again. The error is `Cannot read property 'writeDeprecateLine' of undefined`, which implies that `this.ui` is undefined, not the function itself. This still causes the same error because you removed my check of `this.ui` seperately again.

kellyselden:
My apologies. I tweaked your code to match some of the existing code in the file as far as ui checking. I will fix.

goldenice:
No problem :+1: Happens to the best of us :wink:

kellyselden:
How is 5.2.7?

goldenice:
Works like a charm :+1: Thanks for your quick follow-up, I really appreciate it :smile:

PR #217, opened 2018-04-11
Bumps `broccoli-funnel` version

Bartheleway:
This should remove the warning with the `exists-sync` deprecated dependency (https://github.com/broccolijs/broccoli-funnel/pull/100)


rwjblue:
Thank you!

PR #218, opened 2018-05-18
WIP: Avoid eagerly requiring plugins/presets.

rwjblue:
This is a quick spike to enable parallel transpilation with our default configuration.

/cc @stefanpenner @mikrostew

### DIFF
--- a/index.js
+++ b/index.js
@@ -224,9 +224,12 @@ module.exports = {
       this._getDebugMacroPlugins(config),
       this._getEmberModulesAPIPolyfill(config),
       shouldCompileModules && this._getModulesPlugin(),
-      shouldRunPresetEnv && this._getPresetEnvPlugins(addonProvidedConfig),
       userPostTransformPlugins
     ).filter(Boolean);
+
+    options.presets = [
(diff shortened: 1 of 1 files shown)


rwjblue:
The primary issue with _not_ `require`ing these plugins directly is that they will not be resolved relative to ember-cli-babel itself (instead they will be resolved from the top-level node_modules dir as if they were direct project deps). This would break having multiple independent versions of ember-cli-babel.

rwjblue:
I believe that we can fix that by using `resolve.sync` to resolve the full paths relative to ember-cli-babel, and that `babel` itself allows you to specify a fully qualified path to require as the plugin name.

stefanpenner:
@rwjblue i think we can use https://github.com/babel/broccoli-babel-transpiler#requirefile-required + `require.resolve`, right? cc @mikrostew

stefanpenner:
Also, complementary PR: https://github.com/babel/broccoli-babel-transpiler/pull/137

mikrostew:
To avoid breaking multiple version of ember-cli-babel, it looks to me that we can use `require.resolve` to get the full path as seen from ember-cli-babel, and pass that full path to babel: https://babeljs.io/docs/plugins/#pluginpreset-paths

PR #220, opened 2018-05-21
upgrade deps

stefanpenner:


package.json
```diff
@@ -29,48 +29,53 @@
...
+    "babel-preset-env": "^1.7.0",
+    "broccoli-babel-transpiler": "^6.3.0",
+    "broccoli-debug": "^0.6.4",
     "broccoli-funnel": "^2.0.0",
     "broccoli-source": "^1.1.0",
     "clone": "^2.0.0",
-    "ember-cli-version-checker": "^2.1.0",
-    "semver": "^5.4.1"
+    "ember-cli-version-checker": "^2.1.2",
+    "ember-source-channel-url": "^1.1.0",
```
josemarluedke: Isn't this dependency only used for development mode? Is it necessary to keep it under `dependencies` or could it be moved to `devDependencies`?

stefanpenner: oops, thanks.

stefanpenner: https://github.com/babel/ember-cli-babel/pull/221

PR #223, opened 2018-05-23
Avoid eagerly requiring plugins and presets

mikrostew:
This is a working version of https://github.com/babel/ember-cli-babel/pull/218, to enable parallel transpilation with the default configuration.

This adds the use of `require.resolve` to enable multiple independent versions of ember-cli-babel to coexist, and fixes the tests to account for the changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 ---
 language: node_js
 node_js:
-  - "4"
   - "6"
   - "8"
   - "10"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,10 +8,9 @@ init:
 environment:
   MOCHA_REPORTER: "mocha-appveyor-reporter"
   matrix:
-    - nodejs_version: "4"
     - nodejs_version: "6"
-    - nodejs_version: "7"
-    # - nodejs_version: "8"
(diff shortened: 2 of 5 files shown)


stefanpenner:
I think we should not rely on them aging out in tests

Rather we should likely explicitly kill them for now from the tests (and age out in addition seems fine) but I would hate for my tests to hang (slowing down my own productivity)

The correct solution would be to shutdown workers once the plugin or build pipeline instance that spawned them is cleanedup/destroyed. That may not be practical today (if it is we should doit) So instead, we can likely just manually kill them when the tests shutdown.

stefanpenner:
Some funky windows test failures, I am:

* going to bump the timeout slightly
* consider dropping node 4 testing, since its EOL.

stefanpenner:
released as v6.13.0 🎉

Turbo87:
> consider dropping node 4 testing, since its EOL.

☝️ that was a breaking change!

rwjblue:
:sob: we need to revert https://github.com/babel/ember-cli-babel/commit/e33466e0ee82f564de6e1f410b35d6b734e143fb at least (and add 10 manually)...

PR #224, opened 2018-05-26
Fix presets array bug

twokul:
In `ember-cli` when using `DELAYED_TRANSPILATION` feature, we explicitly
set `disablePresetEnv` flag to `true`. `shouldRunPresetEnv` is a
negation of `disablePresetEnv`.

Prior to this change `options.presets` was an array with one `false`
item in it (`shouldRunPresetEnv` was set to `false`) and would not let files
be transplied when `DELAYED_TRANSPILATION` was enabled.

Useful links:

+ https://github.com/ember-cli/ember-cli/blob/2e0cade64c4698cc48b1fdc20fda2219e63cc973/lib/models/addon.js#L256

### DIFF
--- a/index.js
+++ b/index.js
@@ -223,7 +223,7 @@ module.exports = {
 
     options.presets = [
       shouldRunPresetEnv && this._getPresetEnvPlugins(addonProvidedConfig),
-    ]
+    ].filter(Boolean);
 
     if (shouldCompileModules) {
       options.moduleIds = true;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -221,9 +221,13 @@ module.exports = {
       userPostTransformPlugins
     ).filter(Boolean);
 
-    options.presets = [
-      shouldRunPresetEnv && this._getPresetEnvPlugins(addonProvidedConfig),
-    ]
```
rwjblue: Would it be easier to add `.filter(Boolean)` here instead?

twokul: I think the condition is easier to read than `&&`

twokul: peer pressure 😝

rwjblue:
I’ll get it released in an hour or so...

kellyselden:
Released in v6.14.1

PR #230, opened 2018-07-11
update babel-plugin-debug-macros to 0.2.0-beta.6

kellyselden:
Mentioned here https://github.com/chadhietala/babel-plugin-debug-macros/pull/69#issuecomment-403491412

### DIFF
--- a/index.js
+++ b/index.js
@@ -244,16 +244,19 @@ module.exports = {
     const isProduction = process.env.EMBER_ENV === 'production';
 
     let options = {
-      envFlags: {
-        source: '@glimmer/env',
-        flags: { DEBUG: !isProduction, CI: !!process.env.CI }
-      },
+      flags: [
+        {
+          source: '@glimmer/env',
+          flags: { DEBUG: !isProduction, CI: !!process.env.CI }
+        }
+      ],
 
       externalizeHelpers: {
         global: 'Ember'
       },
 
       debugTools: {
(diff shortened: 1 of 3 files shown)


rwjblue:
[CI passes](https://travis-ci.org/babel/ember-cli-babel/jobs/403042766) (which is great!), but emits a bunch of warnings to the console like this:

```
babel-plugin-debug-macros configuration API has changed, please update your configuration
```

To fix those warnings we need to tweak this section:

https://github.com/babel/ember-cli-babel/blob/8d8c5919f123073289ed31ac567e4dba8f8abddf/index.js#L246-L260

You can compare the ["legacy config" test](https://github.com/chadhietala/babel-plugin-debug-macros/blob/e23b7b39d7746469e25f1232746a78b351128314/tests/create-tests.js#L165-L196) with [this test](https://github.com/chadhietala/babel-plugin-debug-macros/blob/master/tests/create-tests.js#L198-L220) in babel-plugin-debug-macros for an example.

I think we need something like:

```js
let options = {
  externalizeHelpers: {
    global: "Ember"
  },
  debugTools: {
    isDebug: !isProduction,
    source: "@ember/debug",
    assertPredicateIndex: 1
  },
  flags: [
    {
      source: "@glimmer/env",
      flags: { DEBUG: !isProduction, CI: !!process.env.CI }
    }
  ]
};
```

PR #234, opened 2018-08-03
Update to work with latest Babel 7 beta

### LINKED ISSUE #226: Module compilation in 7.0.0-beta.2 includes the full file paths
If I upgrade to `ember-cli-babel@7.0.0-beta.2` in a fresh app (generated from `ember-cli@3.1.4`), the module names for everything under `app` include the full path on disk where they originated, e.g.

```js
define("/Users/dfreeman/Desktop/Scratch Code/babel-7-app/babel-7-app/app", ["exports", "/Users/dfreeman/Desktop/Scratch Code/babel-7-app/babel-7-app/resolver", "ember-load-initializers", "/Users/dfreeman/Desktop/Scratch Code/babel-7-app/babel-7-app/config/environment"], function (_exports, _resolver, _emberLoadInitializers, _environment) {
  "use strict";
  // ...
```

Presumably module names should match those generated by `ember-cli-babel@6.x`?

dfreeman:
This should fix #226 and get things in sync with the most recent Babel 7 beta. A few changes of note:
 - the various `@babel` packages no longer support node 4, so I updated the `engines` field and Travis/Appveyor config here to match what I believe is the standard support matrix for new addons today
 - `@babel/preset-env` is now way pickier about what config it receives, so we need to whitelist the keys we pass in
 - https://github.com/babel/babel/pull/8044 means all module paths are now absolute, so our config for Babel itself and `amd-name-resolver` needed to be tweaked to account for that
 - the latest `babel-plugin-debug-macros` has a new preferred config format

/cc @rwjblue

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 ---
 language: node_js
 node_js:
-  - "4"
   - "6"
-  - "7"
   - "8"
+  - "10"
 
 sudo: false
 
@@ -15,7 +14,6 @@ matrix:
   fast_finish: true
   allow_failures:
     - env: EMBER_TRY_SCENARIO=ember-canary
-    - node_js: "8"
 
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash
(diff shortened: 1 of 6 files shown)


rwjblue:
Awesome, thank you!

PR #237, opened 2018-08-04
Use correct preset env...

rwjblue:


dfreeman:
👍I have a bit of follow-on work to this locally that should get things here in good shape again, but it'll need https://github.com/babel/broccoli-babel-transpiler/pull/148 — I'm working on getting that test suite green in the meantime

PR #236, opened 2018-08-03
feat: make processed file extensions configurable

buschtoens:
This adds an optional `extensions` property to the `ember-cli-babel` config hash and allows users to process files that don't end in `.js`.

For instance:

```js
  let app = new EmberApp(defaults, {
    'ember-cli-babel': {
      extensions: ['js', 'ts']
    },
    babel: {
      plugins: [
        '@babel/transform-typescript',
        '@babel/proposal-class-properties',
        '@babel/proposal-object-rest-spread'
      ]
    }
  });
```

I branched off of `v7.0.0-beta.3`, because the latest commit bricked the `package.json` by referring to a package on the file system. I'll rebase, when this is fixed.

PR #238, opened 2018-08-05
fix: require @babel/preset-env instead of babel-preset-env

buschtoens:
We're listing `@babel/preset-env` as a dependency and so should be using it. 😄 

This was the reason for transpilation failing on me for IE11, because some old incompatible transforms are used by `babel-preset-env`.

AFAICT I can now fully build our TypeScript app with _just_ ember-cli-babel. 😍 🎉 

Simple demo repo here: https://github.com/buschtoens/ember-typescript-babel-7

Thank you guys *so much*. ❤️ 

Closes #236, because it branched off of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ interface EmberCLIBabelConfig {
     disableDebugTooling?: boolean;
     disablePresetEnv?: boolean;
     disableEmberModulesAPIPolyfill?: boolean;
+    extensions?: string[];
   };
 }
 ```
--- a/index.js
+++ b/index.js
@@ -60,7 +60,7 @@ module.exports = {
   setupPreprocessorRegistry(type, registry) {
     registry.add('js', {
       name: 'ember-cli-babel',
-      ext: 'js',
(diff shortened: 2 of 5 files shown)


buschtoens:
One more thing: I added `@babel/core@^7.0.0-beta.56` as a dev dep. But it should probably be a real dependency, right? Because right now the host app needs to install `@babel/core` for us. Is this intentional?

rwjblue:
> Because right now the host app needs to install @babel/core for us.

No, `broccoli-babel-transpiler` has a direct dependency on `@babel/core`.

rwjblue:
Can you rebase (@dfreeman got CI passing in #239)?

buschtoens:
> No, broccoli-babel-transpiler has a direct dependency on @babel/core.

True! Also noticed that. The problem for me was, that some of my transforms had a peer dep on `^beta.56`, but `broccoli-babel-transpiler` has `42` or something like that.

So I guess this PR is outdated now. I'll rebase #236 instead.

PR #239, opened 2018-08-05
Get tests passing with Babel 7 and throwUnlessParallelizable

dfreeman:
This ensures our custom module path de-absolute-izing cooperates with broccoli-babel-transpiler's parallel API, which should get tests passing again.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,8 @@
   "license": "MIT",
   "author": "Gordon Kristan",
   "files": [
-    "index.js"
+    "index.js",
+    "lib"
   ],
   "directories": {
     "doc": "doc",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -37,15 +38,16 @@
     "test:all": "mocha node-tests && ember try:each"
   },
   "dependencies": {
-    "@babel/plugin-transform-modules-amd": "^7.0.0-beta.55",
-    "@babel/preset-env": "^7.0.0-beta.55",
+    "@babel/core": "^7.0.0-beta.56",
```
rwjblue: Seems slightly odd to force the `@babel/core` dep here. Was it to prevent peer dep warnings?

dfreeman: Nope, it was actually because I was getting a `Cannot find module: '@babel/core'`. I'd assumed it was because some last transitive dep had finally shifted it from a dep to a peer dep, but looking at the tree now I think it's more likely that my `node_modules` was just in a weird place because yarn isn't super good at `git` dependencies...

The peerDep warnings _are_ super annoying, but weren't the motivator for this. Happy to revert now that I know it's not necessary.

PR #251, opened 2018-11-08
Add ability to deduplicate babel helpers.

pzuraq:
This PR sets up the groundwork for allowing Ember apps to externalize Babel helpers, reducing the amount of code shipped to the client by deduplicating common code included by Babel.

Externalized helpers represent a unique problem, because in order to truly deduplicate all helpers, the app _and_ all of its addons need to have a shared contract for which helpers will be included in the global namespaces and which will not. Currently we don't have a method to coalesce configurations from many addons, so the current implementation only works with a single, global, app-wide configuration. Eventually this could be set to reasonable defaults, but now it is left off.

Still WIP, definitely needs testing and tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,7 @@ interface EmberCLIBabelConfig {
   */
   'ember-cli-babel'?: {
     includePolyfill?: boolean;
+    includeExternalHelpers?: boolean;
     compileModules?: boolean;
     disableDebugTooling?: boolean;
     disablePresetEnv?: boolean;
@@ -155,6 +156,36 @@ To include it in your app, pass `includePolyfill: true` in your `ember-cli-babel
 let app = new EmberApp(defaults, {
   'ember-cli-babel': {
     includePolyfill: true
+  }
+});
+```
+
+#### External Helpers
+
+Babel often includes helper functions to handle some of the more complex logic
+in codemods. These functions are inlined by default, so they are duplicated in
+every file that they are used in, which adds some extra weight to final builds.
+
+Enabling `includeExternalHelpers` will cause Babel to import these helpers from
+a shared module, reducing app size overall. This option is available _only_ to
+the root application, because it is a global configuration value due to the fact
+that there can only be one version of helpers included.
+
+Note that there is currently no way to whitelist or blacklist helpers, so all
+helpers will be included, even ones which are not used. If your app is small,
+this could add to overall build size, so be sure to check.
+
+`ember-cli-babel` will attempt to include helpers if it believes that it will
+lower your build size, using a number of heuristics. You can override this to
+force inclusion or exclusion of helpers in your app by passing `true` or `false`
+to `includeExternalHelpers` in your `ember-cli-babel` options.
+
+```js
+// ember-cli-build.js
+
+let app = new EmberApp(defaults, {
+  'ember-cli-babel': {
+    includeExternalHelpers: true
   }
 });
 ```
--- a/index.js
+++ b/index.js
@@ -5,6 +5,11 @@ const clone = require('clone');
 const path = require('path');
 const semver = require('semver');
 
+const defaultShouldIncludeHelpers = require('./lib/default-should-include-helpers');
+const findApp = require('./lib/find-app');
+
+const APP_BABEL_RUNTIME_VERSION = new WeakMap();
+
 let count = 0;
 
 module.exports = {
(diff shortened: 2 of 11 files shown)


index.js
```diff
@@ -84,18 +84,88 @@ module.exports = {
     }
   },
 
+  _shouldIncludeHelpers() {
+    let appOptions = (this.app && this.app.options) || {};
```
rwjblue: I'm concerned with this...

`this.app` will only be present for top level addons (direct dependencies of the project), but including or not include helper should be a global thing. As it stands, we will only ever return `true` for direct project deps and any nested addons would continue to embed their helpers...

index.js
```diff
@@ -84,18 +84,88 @@ module.exports = {
...
+      return customOptions.includeExternalHelpers === true;
+    } else {
+      return false;
+    }
+  },
+
+  _importHelpers(app) {
+    let helpersPath = 'vendor/babel-helpers/babel-helpers.js';
+
+    if (this.import) {  // support for ember-cli >= 2.7
```
rwjblue: ember-cli-babel@7 explicitly only supports ember-cli@2.13 and higher (also documented in the README), these conditionals can be removed...

index.js
```diff
@@ -84,18 +84,88 @@ module.exports = {
...
+      let polyfillTree = new Funnel(new UnwatchedDir(polyfillDir), {
+        destDir: 'babel-polyfill'
+      });
+
+      trees.push(polyfillTree);
+    }
+
+    // Helpers are a global config, so only the root application should bother
+    // generating and including the file.
+    if (this.parent === this.app && this._shouldIncludeHelpers()) {
```
rwjblue: Is this ever true? For top level addons I would expect the check to be `this.parent === this.project`...

index.js
```diff
@@ -84,18 +84,88 @@ module.exports = {
...
+
+      let helpers = babel.buildExternalHelpers(this._getHelperWhitelist);
+      let babelHelpersTree = writeFile('babel-helpers/babel-helpers.js', helpers);
+
+      trees.push(babelHelpersTree);
+    }
+
+    if (trees.length > 0) {
+      return new MergeTrees(trees);
+    }
```
rwjblue: Don't we need an `else`?

index.js
```diff
@@ -188,6 +270,7 @@ module.exports = {
     let userPostTransformPlugins = addonProvidedConfig.postTransformPlugins;
 
     options.plugins = [].concat(
+      shouldIncludeHelpers ** this._getHelpersPlugin(),
```
rwjblue: What does this do? I've not seen `**` used like this before...

pzuraq: Thaaaat's supposed to be `&&`, this is what I get for submitting code just before bed haha 😅

index.js
pzuraq: You are correct, that is what I was going for though basically

index.js
pzuraq: the previous early return did `return undefined` so I assumed the return would be optional here

rwjblue: ah, gotcha. yeah thats totally fine

pzuraq:
Alright, I think this is ready for another round of review. Added tests and made sure everything actually working the way it should be!

README.md
```diff
@@ -159,6 +160,33 @@ let app = new EmberApp(defaults, {
 });
 ```
 
+#### External Helpers
+
+Babel often includes helper functions to handle some of the more complex logic
+in codemods. These functions are inlined by default, so they are duplicated in
```
mike-north:
```suggestion
in codemods. These functions are inlined at each location of use by default, so they are duplicated many times, across an application.
```

README.md
```diff
@@ -159,6 +160,33 @@ let app = new EmberApp(defaults, {
...
+every file that they are used in, which adds some extra weight to final builds.
```
mike-north:
```suggestion
Understandably, this adds some extra weight to final builds, scaling up in proportion to use of the related language features (decorators, for...of loops, object or array spread operator, etc...).
```

README.md
```diff
@@ -159,6 +160,33 @@ let app = new EmberApp(defaults, {
...
+
+Enabling `includeExternalHelpers` will cause Babel to import these helpers from
```
mike-north:
```suggestion
Enabling `includeExternalHelpers` will cause Babel to create a shared module for these helpers
```

README.md
```diff
@@ -159,6 +160,33 @@ let app = new EmberApp(defaults, {
...
+a shared module, reducing app size overall. This option is available _only_ to
```
mike-north:
```suggestion
and import them at each call site, reducing the download/parse cost of these language features significantly. This option is available _only_ to
```

index.js
```diff
@@ -123,6 +191,12 @@ module.exports = {
     return (this.parent && this.parent.options) || (this.app && this.app.options) || {};
   },
 
+  _getHostOptions() {
+    let host = this._findHost();
```
dfreeman: The `EngineAddon` wrapper in ember-engines [overrides `_findHost`](https://github.com/ember-engines/ember-engines/blob/master/lib/engine-addon.js#L439) to point to the first lazy-loading engine it comes to rather than the root app. We may need to explicitly traverse the `parent` chain ourselves here.

dfreeman: Actually, since this only ever needs the host options when `this.parent === this.project`, I think we can just access `this.app` directly and not need to bother with `findHost`.

index.js
```diff
@@ -84,19 +86,85 @@ module.exports = {
...
+
+    return this._helperVersion;
+  },
+
+  _getHelpersPlugin() {
+    return [
+      [
+        require.resolve('@babel/plugin-transform-runtime'),
+        {
+          version: this._getHelperVersion(),
```
dfreeman: I see how they're using `version` in the implementation, but I can't find any documentation for it on [the plugin page](https://babeljs.io/docs/en/babel-plugin-transform-runtime). I don't imagine that's intended to be a "secret" option, but have you seen anything indicating it's public?

pzuraq: I have not, but it's definitely needed to actually include decorator helpers or anything more recent than 7.0.0-0, so I can't imagine it's not public? Does seem very strange to me though

PR #260, opened 2019-01-09
Update `@ember/string` detection to check if it is the project being run

locks:
When running the tests for `@ember/string`, `ember-cli-babel` does not detect it properly and tries to process the imports for that module, leading to a failed test run.


locks:
@rwjblue 🤦‍♂️ updated.

Issue #263, opened 2019-01-28
@ember/jquery detection seems off 


rwjblue:
After digging into https://github.com/emberjs/ember-test-helpers/issues/545 a bit, I believe that the code we use to detect if ember-cli-babel's direct parent includes `@ember/jquery` as a dependency is a bit off (note this _may_ also apply to `@ember/string`, tbd).

Specifically, we use `ember-cli-version-checker` to check the version of the `@ember/jquery` instance available from `this.parent.root`. Internally `ember-cli-version-checker` is using standard node resolution algorithm to determine the version that would be required from that root path. Since both yarn and npm@3+ use as flat of a heirarchy as possible, this means that if _anyone_ depends on `@ember/jquery` throughout the system or if that module **happened** to be present (from a prior ember-try scenario perhaps 😸) there will be a `node_modules/@ember/jquery/package.json` found. I believe this is ultimately causing the bug reported in https://github.com/emberjs/ember-test-helpers/issues/545...

Path forward time, I think that we should:

Check `'@ember/jquery' in this.parent.dependencies()` (present on both addon and project base classes) before checking for the version. If `@ember/jquery` is *not* in the parents dependencies, we would not be blacklisting.

@simonihmig - Thoughts?
@locks 	- Can you poke at the `@ember/string` side to see if this is also an issue there?

PR #264, opened 2019-01-28
Only blacklist jQuery module when parent depends on @ember/jquery itself

rwjblue:
Fixes https://github.com/babel/ember-cli-babel/issues/263

### DIFF
--- a/index.js
+++ b/index.js
@@ -427,7 +427,7 @@ module.exports = {
       ];
     }
 
-    if (this._emberJqueryDependencyPresent()) {
+    if (this._shouldBlacklistJQuery()) {
       blacklist['jquery'] = ['default'];
     }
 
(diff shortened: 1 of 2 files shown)


rwjblue:
@simonihmig - Would you mind sanity checking this one?

rwjblue:
> But this would not explain the error reported in emberjs/ember-test-helpers#545, as at least one addon would have @ember/jquery as a dependency, so the jQuery module shim would be available, right?

Yes, I'm waiting for a concrete reproduction, but I was able to reproduce the failure by having done `npm install @ember/jquery && git checkout package.json` (basically installing the addon into node_modules, but not listing it in package.json).

> If we apply the changes here, this could lead to a situation, where a couple of addons might effectively rely on jQuery, but only those that had been updated to list @ember/jquery as a dependency would import jQuery form the shim, and the others would still use ember.$, thus triggering a deprecation, right? Which would work, and maybe even be useful to nudge those addons to list @ember/jquery as a dependency.

Yes, exactly.

> But on the other hand, if there are unmaintained addons, a user would not be able to quash the deprecation or even update to Ember 4 by just installing @ember/jquery in the app itself, which would be possible right now!

I agree, this seems non-ideal. 🤔 I think we should address this by also checking project deps, what do you think?

rwjblue:
@simonihmig I just pushed another commit with that change, what do you think?

simonihmig:
> I think we should address this by also checking project deps, what do you think?

Yes, I think that makes sense! 👍

PR #265, opened 2019-01-29
Fix `@ember/string` detection

locks:
`@ember/string` imports should not be transformed if the addon is a transitive dependency only.


locks:
@rwjblue ready for review!

PR #268, opened 2019-02-14
Update babel-plugin-debug-macros

jrjohnson:
Dropping the beta tag and updating to the latest release version

This release dropped support for node 4, but that was dropped here already.


rwjblue:
Thank you!

PR #271, opened 2019-02-26
chore: adds homepage field to pkg.json

jasonmit:
Minor but enhances the experience within tools like vscode for quick linking to the project from within pkg.json


rwjblue:
Thank you!!

PR #275, opened 2019-03-28
[BUGFIX] Dedupe internal decorators plugins

pzuraq:
babel-plugin-filter-imports currently adds the internal
`decorators` plugin so it can parse and remove decorators,
but this conflicts with the _actual_ decorators plugin. This
PR dedupes the plugins for now (and for previous users of
babel-plugin-filter-imports), we'll also release a new major
version of that plugin that doesn't add the internal plugin.

Related: https://github.com/ember-cli/babel-plugin-filter-imports/pull/135

### DIFF
--- a/index.js
+++ b/index.js
@@ -345,6 +345,14 @@ module.exports = {
       plugins.push([require.resolve('@babel/plugin-proposal-class-properties'), { loose: true }]);
     }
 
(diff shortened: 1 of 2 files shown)


rwjblue:
ember-cli-babel 7.7.1 published 🎉

PR #276, opened 2019-03-28
[BUGFIX] Ensure class properties/decorator plugins are added in correct order

pzuraq:
This PR uses the updated babel-plugin-helpers library to actually modify
the user's plugins and add the decorator/class properties transforms
relative to each other, if one shouldn't exist. They are also
constrained to come _before_ the Typescript transforms, should they
exist. This allows TS to react to any of the output of the previous
transforms.

### DIFF
--- a/index.js
+++ b/index.js
@@ -292,10 +292,13 @@ module.exports = {
     let userPlugins = addonProvidedConfig.plugins;
     let userPostTransformPlugins = addonProvidedConfig.postTransformPlugins;
 
+    if (shouldIncludeDecoratorPlugins) {
+      userPlugins = this._addDecoratorPlugins(userPlugins.slice());
+    }
+
     options.plugins = [].concat(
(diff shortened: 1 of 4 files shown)


rwjblue:
```
 1) ember-cli-babel
       _getDecoratorPlugins
         should include babel transforms by default:
     TypeError: this.addon._getDecoratorPlugins is not a function
      at Context.<anonymous> (node-tests/addon-test.js:689:25)
  2) ember-cli-babel
       _getDecoratorPlugins
         should not include babel transforms if it detects decorators plugin:
     TypeError: this.addon._getDecoratorPlugins is not a function
      at Context.<anonymous> (node-tests/addon-test.js:699:25)
  3) ember-cli-babel
       _getDecoratorPlugins
         should not include babel transforms if it detects class fields plugin:
     TypeError: this.addon._getDecoratorPlugins is not a function
      at Context.<anonymous> (node-tests/addon-test.js:715:25)

```

PR #282, opened 2019-06-11
pin @babel/preset-env to 7.4.2 as 7.4.3 appears to fail

stefanpenner:


rwjblue:
I spent a bit of time debugging this with @chadhietala a few weeks ago, this is an issue with not having a `config/targets.js` setup **and** using `@babel/preset-env@7.4.3` which adds 3 new babel plugins that are required for IE9 and IE10.

The _ultimate_ culprit is babel-plugin-ember-modules-api-polyfill (its emitted AST is invalid, as it mutates an identifier's value to include a path which is completely incorrect).

PR #283, opened 2019-06-12
Revert "Update dependencies"

rwjblue:
This reverts commit f929e209ffc80cfdfdd022438cc3fcb196733985.

This commit was pushed directly to master and broke CI.


stefanpenner:
We run our CI with `--no-lockfile` already, the upgrades here break local development, but still keep CI red.

PR #286, opened 2019-06-13
Ensure `@ember/application/deprecations` is stripped in addition to `@ember/debug`.

bekzod:
https://github.com/emberjs/ember.js/pull/17926#issuecomment-484987305

### DIFF
--- a/index.js
+++ b/index.js
@@ -383,45 +383,52 @@ module.exports = {
   _getDebugMacroPlugins(config) {
     let addonOptions = config['ember-cli-babel'] || {};
 
-    if (addonOptions.disableDebugTooling) { return; }
+    if (addonOptions.disableDebugTooling) {
+      return;
+    }
 
     const isProduction = process.env.EMBER_ENV === 'production';
     const isDebug = !isProduction;
 
-    let options1 = {
-      flags: [
-        {
-          source: '@glimmer/env',
-          flags: { DEBUG: isDebug, CI: !!process.env.CI }
-        }
-      ],
-
-      externalizeHelpers: {
-        global: 'Ember'
-      },
-
-      debugTools: {
-        isDebug,
-        source: '@ember/debug',
-        assertPredicateIndex: 1
-      }
-    };
-
-    let options2 = { // deprecated import path https://github.com/emberjs/ember.js/pull/17926#issuecomment-484987305
-      externalizeHelpers: {
-        global: 'Ember'
-      },
-
-      debugTools: {
-        isDebug,
-        source: '@ember/application/deprecations',
-        assertPredicateIndex: 1
-      }
-    };
-
     return [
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -407,7 +407,22 @@ module.exports = {
...
+
+      debugTools: {
+        isDebug,
+        source: '@ember/application/deprecations',
+        assertPredicateIndex: 1
+      }
+    };
+
+    return [
+      [require.resolve('babel-plugin-debug-macros'), options1],
```
rwjblue:
```suggestion
      [require.resolve('babel-plugin-debug-macros'), options1, '@ember/debug stripping'],
```

index.js
```diff
@@ -407,7 +407,22 @@ module.exports = {
...
+      [require.resolve('babel-plugin-debug-macros'), options2, 'remove deprecated debug']
```
rwjblue:
```suggestion
      [require.resolve('babel-plugin-debug-macros'), options2, '@ember/application/deprecations stripping']
```

bekzod:
thank you @rwjblue, accepted all those changes

rwjblue:
@bekzod - I've pushed a few changes:

* added a failing test that attempts to confirm the babel-plugin-ember-modules-api-polyfill doesn't transpile `import { deprecate } from '@ember/application/deprecations'` to globals (e.g. `Ember.deprecate`)
* fix that failing test (by ensuring that `deprecate` from `@ember/application/deprecations` is not transformed by the modules polyfill
* added a test confirming that `import { deprecate } from '@ember/application/deprecations` is transpiled properly in debug builds

Should be good to land once CI is finished...

index.js
```diff
@@ -383,31 +383,53 @@ module.exports = {
...
-      debugTools: {
-        isDebug,
-        source: '@ember/debug',
-        assertPredicateIndex: 1
-      }
-    };
-
-    return [[require.resolve('babel-plugin-debug-macros'), options]];
+      [
+        require.resolve('babel-plugin-debug-macros'),
```
stefanpenner: running the same plugin twice seems unfortunate, can the primary plugin simply be taught how to handle this case?

rwjblue: We already support arbitrary module paths with `flags`, but not with `debugTools`. We can definitely update `babel-plugin-debug-macros` to support that though.

stefanpenner: i think this may be tripping up https://github.com/babel/babel/blob/11fa2461cec930168fd43cfeda733aab056d8318/packages/babel-core/src/config/config-descriptors.js#L240

stefanpenner: investigating

stefanpenner: The issue seems to have been a monorepo with a shared lockfile resulted in an incomplete semver drift of  relevant dependencies, upgrading all repos contained within the monorepo fixed the issues.

PR #291, opened 2019-07-09
upgrade all in-range deps

NullVoxPopuli:
The babel update should get us the nice generator syntax back.

```ts
@task *myTask() {

}
```

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -10,13 +10,13 @@
     "@babel/highlight" "^7.0.0"
 
 "@babel/core@^7.0.0":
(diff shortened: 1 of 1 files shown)


rwjblue:
@NullVoxPopuli - This doesn't really change anything for consumers, it just updates our own lock file. I think its totally fine (though we should update again because 7.5.4 is current babel).

NullVoxPopuli:
@rwjblue do you think we should bump the min versions?

rwjblue:
Nah, not yet.

PR #306, opened 2019-11-11
Respect babel "ignore" option

### LINKED ISSUE #305: How to specify direct babel options (not preset-env options)?
I am trying to set up the [`ignore`](https://babeljs.io/docs/en/options#ignore) option in my babel configuration, but this can only be specified at a root level and is not accepted in the preset env options.

Is there any way for me to provide root options to be forwarded to babel?

monovertex:
In Babel 7 the `ignore` option is no longer valid for @babel/preset-env, so there is currently no way of providing this option through ember-cli-babel.

I followed #158, which added `sourceMaps` option forwarding - this PR is pretty much the same.

Food for thought: should this plugin maybe allow the user to specify options separately for babel and for the env preset, instead of forwarding these manually? Since Babel separates them, it feels a bit weird that this plugin will bundle them together, even if it abstracts the env preset usage.

Fixes #305

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ interface EmberCLIBabelConfig {
     exclude?: string[];
     useBuiltIns?: boolean;
     sourceMaps?: boolean | "inline" | "both";
+    ignore?: MatchPattern[];
     plugins?: BabelPlugin[];
   };
 
--- a/index.js
+++ b/index.js
@@ -280,11 +280,17 @@ module.exports = {
       sourceMaps = config.babel.sourceMaps;
     }
 
+    let ignore;
+    if (config.babel && 'ignore' in config.babel) {
+      ignore = config.babel.ignore;
+    }
+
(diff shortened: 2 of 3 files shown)


monovertex:
I'm not exactly sure why the Travis build fails 🤔

houfeng0923:
ye, we need `ignore`. please

houfeng0923:
@monovertex , if you use `ember-auto-import` , you can try `skipBabel` option .
i resolved my problem with it .

monovertex:
We actually needed this behavior for internal files, but we realized after opening this PR that the files were not ignored with Babel 6 either, the `ignore` option was, well, ignored. The original intention was to not break some sensitive code that was already ES5 and lower compliant, but they were transpiled anyway and they worked in production for quite a while.

We also realized that ignore doesn't really make sense for internal files, because then you'd have to shim the module scaffolding.

I forgot about this PR, so I will close it now, seeing that this is not really needed.

PR #307, opened 2019-11-11
[BUGFIX] Make loose mode in class properties transform respect Babel settings

pzuraq:
Up until recently, loose mode was required to be enabled in class properties if
decorators were also included. This behavior made undecorated class properties
behave unlike the spec, but _decorated_ class properties still did behave like
the spec, counterintuitively. It also meant that we weren't following the user's
settings for loose mode, if they existed.

This PR makes the setting part of the user's standard Babel `loose` setting.

## Possible Impacts

Given the following class definitions:

```js
class Foo {
  get foo() {
    return this._foo;
  }

  set foo(value) {
    this._foo = value;
  }
}

class Bar extends Foo {
  foo = 123;
}
```

With `loose: true`, the subclass's `foo` would activate the superclass's `set foo()`, using it to assign the property on the class. With `loose: false`, the `foo` property will be defined in the subclass using `Object.defineProperty`, which means the getter/setter will effectively be ignored. This could also impact decorated properties that install getters/setters:

```js
class Foo {
  @tracked foo;
}

class Bar extends Foo {
  foo = 123;
}
```

In these cases, strict mode users can use a getter/setter or redecorate the property instead to achieve similar behavior to loose mode.

```js
class Foo {
  @tracked foo;
}

class Bar extends Foo {
  @tracked foo = 123;
}
```

They can also enable loose mode in general, though this is not recommended.


rwjblue:
Looks great, thanks for fixing this upstream! Can you update the description here to make it clearer what happened before your changes, and after? Just so that if folks run into an issue after an ember-cli-babel update they could see what the intended changes here were...

rwjblue:
Looks like the appveyor setup is not using `yarn`, and therefore not getting the pinned resolution we need to support Node 6.

PR #314, opened 2019-12-19
Provide TypeScript compilation support when using ember-cli-typescript@4 or higher.

jamescdavis:
This is based on a [conversation](https://discordapp.com/channels/480462759797063690/484421406659182603/654770983319896064) between some of the `ember-cli-typescript` maintainers (including @dfreeman and @chriskrycho) and @rwjblue. The `ember-cli-typescript` side of this is happening in https://github.com/typed-ember/ember-cli-typescript/pull/1018.

This moves management of `@babel/plugin-transform-typescript` to `ember-cli-babel` when `ember-cli-typescript >= 4.0.0-alpha.1` is installed in the parent. It also adds `@babel/plugin-proposal-optional-chaining` and `@babel/plugin-proposal-nullish-coalescing-operator` when `ember-cli-typescript >= 4.0.0-alpha.1` is installed because these are standard features in TypeScript >= 3.7.0 (and we can't add these conditionally because we can't be sure about the TypeScript version used in addons).

See https://github.com/typed-ember/ember-cli-typescript/pull/1018 for the release plan.


jamescdavis:
The floating deps job has also [failed recently for a docs-only PR](https://travis-ci.org/babel/ember-cli-babel/builds/613929991). ¯\\\_(ツ)\_/¯

rwjblue:
https://github.com/babel/ember-cli-babel/pull/315 should fix the floating deps tests.

index.js
```diff
@@ -254,14 +254,16 @@ module.exports = {
   },
 
   _getExtensions(config) {
+    let shouldHandleTypeScript = this._shouldHandleTypeScript();
     let emberCLIBabelConfig = config['ember-cli-babel'] || {};
-    return emberCLIBabelConfig.extensions || ['js'];
+    return emberCLIBabelConfig.extensions || (shouldHandleTypeScript ? ['js', 'ts'] : ['js']);
```
rwjblue: Any reason to not _always_ default to `['js','ts']`?

index.js
```diff
@@ -320,6 +326,93 @@ module.exports = {
     return options;
   },
 
+  _shouldHandleTypeScript() {
+    if (this._cachedShouldHandleTypescript === undefined) {
+      // TODO: this doesn't account for things like in-repo addons and other
+      // weird inclusion structures
+      let checker = new VersionChecker(this.parent).for('ember-cli-typescript', 'npm');
```
rwjblue: Instead of using `ember-cli-version-checker` here, I think it would probably be better (and may remove that TODO above) to use:

```js
let typescriptAddon = this.parent.addons.find(a => a.name === 'ember-cli-typescript');
return typescriptAddon !== undefined && semver.gte(typescriptAddon.pkg.version, '4.0.0-alpha.1');
```

index.js
```diff
@@ -320,6 +326,93 @@ module.exports = {
...
+  },
+
+	_buildClassFeaturePluginConstraints(constraints) {
+    // With older versions of ember-cli-typescript, class feature plugins like
+    // @babel/plugin-proposal-class-properties were run before the TS transform.
+    // With more recent language features like `declare` field modifiers, Babel
+    // now has assertions requiring that the TS transform runs first, so if we
+    // know we're responsible for setting that transform up, we follow those rules.
+    if (this._shouldHandleTypeScript()) {
+      constraints.after = constraints.after || [];
```
rwjblue: I don't think we need before/after at all since we now control the full list of plugins (we would still need to deal with that in the `else` below though for ember-cli-typescript < 4 support).

index.js
jamescdavis: *probably* no reason not to. If a user wanted to transpile ts some other way for some reason (or ignore ts or something) they could just manually configure the extensions and exclude ts.

index.js
jamescdavis: I'll give it a try!

index.js
jamescdavis: 🤔 yeah, I suppose we only have to force a different order for the exception case (older e-c-ts) since we're just adding them in the correct order to begin with.

jamescdavis: But what about when something *else* has added `@babel/plugin-transform-typescript`? This PR handles this case (but warns) similar to what `ember-cli-babel` does with `@babel/plugin-proposal-decorators` and `@babel/plugin-proposal-class-properties`. I mean, theoretically, if something else has already added `@babel/plugin-transform-typescript`, we'd be adding the class feature plugins after anyways, but, IMHO, it doesn't hurt to be sure.

index.js
rwjblue: Ya, that's what I was thinking. Let's do that.

index.js
jamescdavis: This worked well and had the benefit of being easier to write tests for.

index.js
rwjblue: > But what about when something else has added @babel/plugin-transform-typescript?

1) AFAICT we don't actually have that problem (a search in emberobserver suggests that only the e-c-ts addon does this)
2) I'd rather the code for the _go forward_ world be simpler and clearer. That way when we do a 8.x (or whatever) dropping support for checking for e-c-ts the cleanup is easy.

jamescdavis: 1. I guess I was thinking either something non-Ember or user config.
2. This makes sense. Default to the happy path and just handle the one known issue (pre 4.0 e-c-ts). Anything else is not a supported config (if you're doing something weird, expect breakage).

jamescdavis: I've changed this to only handle the else case (e-c-ts < 4.0). We still need to enforce the decorators before class properties constraint for now because otherwise they get out of order when the class properties plugin is forced to be *before* the ts transform. Once e-c-babel drops support for e-c-ts < 4.0, the calls to `_buildClassFeaturePluginConstraints()` (including the decorators/class properties before/after) can just be removed altogether.

index.js
```diff
@@ -320,13 +325,93 @@ module.exports = {
...
   },
 
+  _shouldHandleTypeScript(config) {
+      let emberCLIBabelConfig = config['ember-cli-babel'] || {};
+      if (typeof emberCLIBabelConfig.enableTypeScriptTransforms === 'boolean') {
+        return emberCLIBabelConfig.enableTypeScriptTransforms;
+      }
+      let typeScriptAddon = this.parent.addons
+        && this.parent.addons.find(a => a.name === 'ember-cli-typescript');
+      return typeof typeScriptAddon !== 'undefined'
```
rwjblue: This definitely works, but I _think_ we can avoid the `typeof` check. Not sure that it is or isn't better though, waht do you think?

```suggestion
      return typeScriptAddon !== undefined
```

index.js
```diff
@@ -320,13 +325,93 @@ module.exports = {
     return options;
   },
 
+  _shouldHandleTypeScript(config) {
+      let emberCLIBabelConfig = config['ember-cli-babel'] || {};
+      if (typeof emberCLIBabelConfig.enableTypeScriptTransforms === 'boolean') {
+        return emberCLIBabelConfig.enableTypeScriptTransforms;
+      }
```
rwjblue: This is great, it allows for a really nice escape hatch when our autodetection may fail (which AFAICT shouldn't happen, but it is still nice to be prepared).

Can you add this to the documentation in the README?

index.js
```diff
@@ -320,13 +325,93 @@ module.exports = {
...
+      let typeScriptAddon = this.parent.addons
+        && this.parent.addons.find(a => a.name === 'ember-cli-typescript');
+      return typeof typeScriptAddon !== 'undefined'
+        && semver.gte(typeScriptAddon.pkg.version, '4.0.0-alpha.0');
+  },
+
+	_buildClassFeaturePluginConstraints(constraints, config) {
```
rwjblue: This is indented with a tab, can you swap to two spaces instead?

```suggestion
  _buildClassFeaturePluginConstraints(constraints, config) {
```

index.js
```diff
@@ -320,13 +325,93 @@ module.exports = {
...
+    // With older versions of ember-cli-typescript, class feature plugins like
```
rwjblue: Can you tweak this comment to indicate specific versions that it applies to? I'm mostly looking to ensure that we can more easily know when the logic can be removed (e.g. after XYZ months from the last release that required it, or X major versions, or something like that but we need the specific versioning to be able to figure that out).

index.js
```diff
@@ -320,13 +325,93 @@ module.exports = {
...
+  },
+
+  _addTypeScriptPlugins(plugins) {
+    const { hasPlugin, addPlugin } = require('ember-cli-babel-plugin-helpers');
+
+    if (hasPlugin(plugins, '@babel/plugin-proposal-optional-chaining')) {
+      if (this.parent === this.project) {
+        this.project.ui.writeWarnLine(`${
+          this._parentName()
+        } has added the optional chaining plugin to its build, but ember-cli-babel provides this by default now when ember-cli-typescript >= 4.0 and typescript >= 3.7 are installed! You can remove the transform, or the addon that provided it.`);
```
rwjblue: Just to double check, this is added by default because TypeScript requires it? Also, how do you know that typescript >= 3.7 is being used (does e-c-ts bring a minimum of typescript@3.7?)?

rwjblue: Side note, we may want to make a more official policy to include things like this in the future (e.g. "stage-3 proposals are added by default"), not sure but that would _probably_ require an RFC. 

What do you think @pzuraq?

index.js
jamescdavis: The `undefined` global used to be able to be redefined, so that check was dangerous, but I guess nothing in the wild allows this anymore: https://kangax.github.io/compat-table/es5/#test-Immutable_globals_undefined

index.js
jamescdavis: Not sure how that happened. Will do!

index.js
jamescdavis: Yes, this is because these features are part of TS >= 3.7. The reason these have to be added unconditionally is because we *don't* know the TypeScript version an addon might have used.

index.js
jamescdavis: There are also people wishing to use TypeScript *without* integrated type-checking, which this option would enable (by not having to install `ember-cli-typescript` just to get TypeScript transpilation).

jamescdavis: I also documented the automatic behavior and incompatibility of this option with e-c-ts < 4.0.

dfreeman:
@rwjblue one thing the typed-ember folks had talked about as part of our v4 release was explicitly excluding `.d.ts` files from emit. Currently, any of those that you might have in your `app`/`addon` tree will each produce an empty module definition with the trailing `.d` (`define('foo/bar/baz.d', ...)`) in the final output, which is a waste of bytes.

Do you have thoughts on how we might best accomplish that?

rwjblue:
@dfreeman - When running our `toTree` we can add a `new Funnel(rawInput, { exclude: ['**/*.d.ts'] })` before actually passing to compilation with `broccoli-babel-transpiler`.

rwjblue:
Tweaking this 

https://github.com/babel/ember-cli-babel/blob/442ec64a737a8cd638ca2fa064643bf258aa8d71/index.js#L44-L58

To be something like:

```js
  transpileTree(inputTree, config) {
    let description = `000${++count}`.slice(-3);
    let postDebugTree = this._debugTree(inputTree, `${description}:input`);

    let options = this.buildBabelOptions(config);
    let output;
    if (this._shouldDoNothing(options)) {
      output = postDebugTree;
    } else {
      let BabelTranspiler = require('broccoli-babel-transpiler');
      
      let transpilationInput = postDebugTree;

      if (this._shouldCompileTypescriptOrWhateverTheMethodIsCalled()) {
        transpilationInput = this._debugTree(new Funnel(postDebugTree, { excludes: ['**/*.d.ts']}), `${description}:filtered-input`);
      }
      output = new BabelTranspiler(transpilationInput, options);
    }

    return this._debugTree(output, `${description}:output`);
  },
```

dfreeman:
👍 Thanks, that makes sense! I wasn't sure if there was a cleverer way than just funneling the input, but I guess that's exactly what we'd be doing if we implemented outside of ember-cli-babel anyway

index.js
jamescdavis: So, turns out this *has* to be conditional. If a package (say, `ember-ajax`) contains `.ts` files and that package isn't using ember-cli-typescript >= 4.0 (say, `ember-ajax`) Babel will choke because it tries to transpile the `.ts` files, but the TypeScript transforms won't be added for Babel in that addon tree. We can only safely tell it to include `.ts` files when we know the Babel instance will have the TS transforms (`shouldHandleTypeScript === true`).

I will revert this back.

jamescdavis:
@rwjblue @dfreeman When I saw [this](https://github.com/babel/broccoli-babel-transpiler/blob/9f58258b7aa9c70afa72b952818c1262a07954e3/index.js#L14), I was hoping there was a way to filter out `.d.ts` files with a negative lookbehind in the extensions list (set extensions to `['js', '(?<!\\.d\\.)ts']`) . But alas, that RegExp is only used to pull module names out of file names (not to filter extensions) 😞. I think it would require modifying both [`ember-cli-preprocess-registry`](https://github.com/ember-cli/ember-cli-preprocess-registry/blob/5e491118b3b2d5de986703c68d1b9874cb733f68/preprocessors.js#L83) and [`broccoli-persistent-filter`](https://github.com/broccolijs/broccoli-persistent-filter/blob/598264ab44fcccdd433abecaacb9181959cdac16/index.js#L388-L392) to handle specifying extensions as regexes.

jamescdavis:
[ember-cli-typescript@4.0.0-alpha.1](https://www.npmjs.com/package/ember-cli-typescript/v/4.0.0-alpha.1) has been released, so this is now ready to be included in an `ember-cli-babel` release when we are satisfied with it. Once released, we will update `ember-cli-typescript` to depend on that version of `ember-cli-babel` (and likely release as `4.0.0-beta.1` or `4.0.0-rc.1` for people to test before we cut a stable release). IMHO, we don't need a pre-release of `ember-cli-babel` for this because the changes will only take effect when combined with a pre-release of `ember-cli-typescript` (moving to a stable release once we're confident we didn't miss something).

jamescdavis:
Since `@babel/preset-env` now [enables optional chaining and nullish-coalescing by default](https://github.com/babel/babel/pull/10811) (as of [7.8.0](https://github.com/babel/babel/releases/tag/v7.8.0)), I've updated this to not add them manually. I've also renamed the method to add the plugin and the flag to manually enable the TypeScript transform to be singular since they now only affect one plugin.

PR #318, opened 2020-01-15
Add polyfill that allows apps and addons to use `@ember-data` related packages regardless of `ember-data` version

runspired:


index.js
```diff
@@ -444,6 +445,16 @@ module.exports = {
     }
   },
 
+  _getEmberDataPackagesPolyfill(config) {
+    let addonOptions = config['ember-cli-babel'] || {};
+
+    if (addonOptions.disableEmberDataPackagesPolyfill) { return; }
```
rwjblue: Please add this config value to the README

index.js
```diff
@@ -511,6 +522,17 @@ module.exports = {
     return true;
   },
 
+  _emberDataVersionRequiresPackagesPolyfill() {
+    let dataAddon = this.project.addonPackages['ember-data'];
```
rwjblue: What is `this.project.addonPackages`? Do you mean `this.project.findAddonByName`?

rwjblue: I think you may also need to check for this transitively though, e.g. if the project itself does not depend on ember-data but some other shared addon does...

runspired: @rwjblue recommended pattern for that detection? similar to what we do in ember-m3?

runspired: or rather what we do in ember-data: https://github.com/emberjs/data/blob/master/packages/private-build-infra/src/packages.js#L5

runspired:
The failing test is for node 6, which I believe we can safely drop.

Indeed this project currently tests 6,8, and 10 when we should bump to testing 10, 12 and 13

index.js
```diff
@@ -511,6 +522,20 @@ module.exports = {
...
 
+  _emberDataVersionRequiresPackagesPolyfill() {
+    let checker = new VersionChecker(this.project);
+    let dep = checker.for('ember-data');
+    let hasEmberData = dep.exists();
+
+    if (hasEmberData) {
+      if (!dep.version) {
+        throw new Error('EmberData missing version');
+      }
```
rwjblue: Do we ever expect this to actually happen? I _think_ no, but just wanted to check and see why it was added...

runspired: When I was creating the test mock environment the failures were pretty obtuse, this made it clear when the issue was no version found so that I knew I needed to continue getting the mock env closer to what version checker needed.

It is removable.

PR #321, opened 2020-01-29
[CHORE BREAKING] bump node support to 10, 12+

runspired:


Turbo87:
this would break lockstep with Babel itself... I'm wondering if there is a real need to do this 🤔

runspired:
Are we in lockstep more than by accident?

Turbo87:
afaik yes

rwjblue:
Hmm, I wouldn't assume we are trying to match babel majors here :thinking:

The thing that _actually_ brings babel is broccoli-babel-transpiler, and it's only 6.x anyways...

bertdeblock:
Closing in favor of #454 if that's okay.

PR #325, opened 2020-02-06
Ensure `ember-data` modules polyfill does not apply to any `@ember-data` packages.

runspired:


index.js
```diff
@@ -511,8 +511,8 @@ module.exports = {
     let addonOptions = config['ember-cli-babel'] || {};
 
     if (addonOptions.disableEmberDataPackagesPolyfill) { return; }
-    // Don't convert ember-data itself!
-    if (this.parent.name === 'ember-data') { return; }
+    // Don't convert ember-data itself or any @ember-data packages!
+    if (this.parent.name === 'ember-data' || this.parent.name.indexOf('@ember-data/') === 0) { return; }
```
rwjblue: Could use `.startsWith` I think (instead of .indexOf() === 0), what do you think?

runspired: @rwjblue I had to add a check that name was actually present. RE startsWith is it compatible with node 6?

runspired: @rwjblue do you think we need to check `parent.name()` as well?

runspired: (I'm guessing no since our own tests would always have a version >= 3.12)

runspired: looks like startsWith has support in node 4 so will use

rwjblue: .name as a function is only going to match the project itself (which means inside ember-data while locally developing), but ember-data can just directly disable the flag in its Babel config. 

Totally up to you if you want to support both, but I don’t think it’s strictly required

runspired: Let’s merge and move on then :)

PR #338, opened 2020-05-29
Update babel-plugin-ember-modules-api-polyfill to fix issues with invalid identifiers.

### LINKED ISSUE #316: Older target browsers break runtime
When setting target browsers in `targets.js` file like this:

```
const browsers = [
  'Android >= 4.1'
];
```

Build succeeds but app actually does no run and fails with error:

```
Uncaught ReferenceError: _EmberObject is not defined
    at Module.callback (index.js:137)
    at Module.exports (loader.js:106)
    at Module._reify (loader.js:143)
    at Module.reify (loader.js:130)
    at Module.exports (loader.js:104)
    at Module._reify (loader.js:143)
    at Module.reify (loader.js:130)
    at Module.exports (loader.js:104)
    at requireModule (loader.js:27)
```

However when I target browser is set to 

```
const browsers = [
  'Android >= 4.4'
];
```

Build succeeds and also app runs perfectly on runtime.

This hints like some sort of Babel transforms mess something up but not sure where is it happening. Is it in this repository, in ember-source or some Babel plugin combinations simply break it.

I have an app in production which has Android >= 4.1 target + Babel v7.12.0 but it's no longer upgradeable as whenever some floating dependency gets bumped it breaks and which of the packages cause this is difficult understand.

**How to reproduce**

Simply create blank addon or app, set target to Android >= 4.1 and it will fail when trying to visit http://localhost:4200

### LINKED ISSUE #322: _EmberApplication is not defined in 7.14.1 build
I just tried updating from ember-cli-babel in my app from `7.13.2` > `7.14.1` and this appears to be breaking my app's  distributable after building.

I am getting 
```
ReferenceError: _EmberApplication is not defined
```
in my console when running the app from a test environment. Runs from `ember serve` livereload appear to not be affected at all.

This happens when this change is the only one between deployments. My installed packages from `package.json` are seen below.

I've already tried nuking `node_modules` for a clean installation and it had no impact.

```json
{
  "scripts": {
    "build": "ember build",
    "lint:hbs": "ember-template-lint .",
    "lint:js": "eslint .",
    "start": "ember serve",
    "test": "ember test",
    "publish": "ember build --environment=production",
    "postpublish": "yarn bestzip dist.zip .ebextensions dist/"
  },
  "devDependencies": {
    "@ember/optional-features": "^1.3.0",
    "@fortawesome/ember-fontawesome": "^0.2.1",
    "@fortawesome/free-brands-svg-icons": "^5.12.0",
    "@fortawesome/free-solid-svg-icons": "^5.12.0",
    "bestzip": "^2.1.5",
    "bootstrap": "^4.4.1",
    "broccoli-asset-rev": "^3.0.0",
    "ember-auto-import": "^1.5.3",
    "ember-bootstrap": "^3.1.1",
    "ember-cli-app-version": "^3.2.0",
    "ember-cli-babel": "7.14.1",
    "ember-cli-cjs-transform": "^2.0.0",
    "ember-cli-dependency-checker": "^3.2.0",
    "ember-cli-deploy": "^1.0.2",
    "ember-cli-deploy-build": "^2.0.0",
    "ember-cli-deploy-compress": "^0.3.2",
    "ember-cli-deploy-gzip": "^1.0.1",
    "ember-cli-eslint": "^5.1.0",
    "ember-cli-htmlbars": "^4.2.2",
    "ember-cli-inject-live-reload": "^2.0.2",
    "ember-cli-sass": "^10.0.1",
    "ember-cli-sri": "^2.1.1",
    "ember-cli-template-lint": "^2.0.0",
    "ember-cli-uglify": "^3.0.0",
    "ember-concurrency": "^1.1.5",
    "ember-cookies": "^0.5.2",
    "ember-export-application-global": "^2.0.1",
    "ember-g-recaptcha": "^1.2.0",
    "ember-intl": "^4.2.3",
    "ember-load-initializers": "^2.1.1",
    "ember-local-storage": "^1.7.1",
    "ember-metrics": "^0.14.0",
    "ember-qunit": "^3.4.1",
    "ember-resolver": "^7.0.0",
    "ember-router-scroll": "^2.0.0",
    "ember-search-helper": "^2.0.0",
    "ember-simple-auth": "^2.1.1",
    "ember-source": "^3.16.0",
    "ember-truth-helpers": "^2.1.0",
    "eslint": "^6.8.0",
    "eslint-plugin-ember": "^7.7.2",
    "loader.js": "^4.7.0",
    "qunit-dom": "^0.9.2",
    "sass": "^1.25.1-test.1"
  },
  "dependencies": {
    "big.js": "^5.2.2",
    "dayjs": "^1.8.19",
    "ember": "^1.0.3",
    "ember-cli": "^3.15.2",
    "query-string": "^6.10.1"
  }
}

### PR DESCRIPTION
**Pull request by rwjblue**
The babel plugin previously took an identifier and replaced it's `name` value with one that included a string like `Ember.Foo`. This is completely invalid (a string like `Ember.Foo` is actually a MemberExpression), but we got away with it because for the most part Babel happily printed **whatever** value you give it in that spot.

Unfortunately, this happy mistake has caused us quite a bit of pain due to Babel introducing new plugins that run as part of `@babel/preset-env` (e.g. to make sure that all identifiers would work properly on IE9), and those plugins **properly** sanitizing invalid identifiers.

The plugin was updated to avoid this mistake, and properly rewrite the identifiers with a member expression. It also added tests to confirm things are working as intended.

Fixes #316
Fixes #322
Fixes https://github.com/emberjs/ember.js/issues/18992

### DIFF
--- a/node-tests/addon-test.js
+++ b/node-tests/addon-test.js
@@ -203,7 +203,7 @@ describe('ember-cli-babel', function() {
 
         expect(contents).to.not.include('@ember/debug');
         expect(contents).to.include('function _asyncToGenerator');
-        expect(contents).to.include('Ember.inspect;');
+        expect(contents).to.include('inspect.call');
         expect(contents).to.not.include('assert');
       }));
     });
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "amd-name-resolver": "^1.2.1",
     "babel-plugin-debug-macros": "^0.3.0",
     "babel-plugin-ember-data-packages-polyfill": "^0.1.2",
-    "babel-plugin-ember-modules-api-polyfill": "^2.13.0",
+    "babel-plugin-ember-modules-api-polyfill": "^2.13.1",
     "babel-plugin-module-resolver": "^3.1.1",
     "broccoli-babel-transpiler": "^7.4.0",
     "broccoli-debug": "^0.6.4",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1574,10 +1574,10 @@ babel-plugin-ember-modules-api-polyfill@^2.12.0, babel-plugin-ember-modules-api-
   dependencies:
     ember-rfc176-data "^0.3.12"
 
-babel-plugin-ember-modules-api-polyfill@^2.13.0:
-  version "2.13.0"
-  resolved "https://registry.yarnpkg.com/babel-plugin-ember-modules-api-polyfill/-/babel-plugin-ember-modules-api-polyfill-2.13.0.tgz#f9d3928d4c40192a44a5b52b67bbdedc4a748ea2"
-  integrity sha512-Q2i7uZMNSuJFYUV8stEsLQIRfMhQJew62sOnR+ESNjb4vlI7Rj1JlSgymTGwmCTVh+poTBpMidiWCoITDtbiIA==
+babel-plugin-ember-modules-api-polyfill@^2.13.1:
+  version "2.13.1"
+  resolved "https://registry.yarnpkg.com/babel-plugin-ember-modules-api-polyfill/-/babel-plugin-ember-modules-api-polyfill-2.13.1.tgz#a529ec260ee2c5af5b5696074b23334535100368"
+  integrity sha512-h9DZ4wD7LYfKK98bZ7Rf6EX7bi5bmfO/hS/jWdqFj4OrpBa4qwRUczBDU7tUlyLcOxyXvXa6+DzXk38I65Y7vA==
   dependencies:
     ember-rfc176-data "^0.3.13"
 

### CODE REVIEW

**Comment by lukemelia** [comment]
Thanks for working on this @rwjblue!

I tried upgraded an addon that was having an issue with this (ember-keyboard), and while my yarn.lock-driven tests pass, my 3.12 ember-try build now fails with a different error:

```
➜  ember-keyboard git:(rfc/api-change) ✗ node_modules/.bin/ember try:one ember-lts-3.12
yarn install v1.22.4
[snip]
Build Error (broccoli-persistent-filter:Babel > [Babel: @glimmer/component]) in @glimmer/component/-private/ember-component-manager.ts

/Users/lmelia/p/ember/ember-keyboard/@glimmer/component/-private/ember-component-manager.ts: Property id of TSDeclareFunction expected node to be of a type ["Identifier"] but instead got "MemberExpression"
```

Let me know if it would be best for me to open a separate issue somewhere, or if I can provide any additional info.

miguelcobain:
The same issue started happening to me as well on two of my apps.

BarryThePenguin:
I'm seeing the same error in a number of apps and addons with `ember-cli-babel@7.20.1`. It looks like a mixed bag, so I'm taking a look to see why some would be failing while others aren't..

rwjblue:
https://github.com/babel/ember-cli-babel/releases/tag/v7.20.2 is out with the fix for the TS issue (thank you @fivetanley!). See https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill/pull/110 for more explanation.

Issue #355, opened 2020-06-15
Replace term `blacklist` with `exclude`


mydea:
Parallel to: https://github.com/emberjs/rfcs/pull/639

I would like to propose to rename the `blacklist` option to `exclude`. It is, as far as I can see (?) only used in private code, so should be possible without any user-facing changes.

As a sidenote, I noticed this by coming from here:
https://ember-cli.com/user-guide/#js-transpiling

Which references this:

```js
 let app = new EmberApp(defaults, {
    babel: {
      blacklist: [
        'es6.arrowFunctions',
        // ...more options
      ]
    }
  });
```

By looking through the code, it doesn't seem this is still supported - or am I misunderstanding this? If so, we can/should probably also remove this from the Ember CLI API docs (which is probably unrelated to this repo itself, so more of a sidenote...)

rwjblue:
@mydea - The API you are referencing above was a Babel 5 API, and is no longer supported. All references to `blacklist` in the current code are private details, and can be changed anytime we want. A PR renaming would be welcome (though I think we'll also need to change [ember-cli/babel-plugin-ember-modules-api-polyfill](https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill)).

rwjblue:
The changes over in babel-plugin-ember-modules-api-polyfill have landed:

https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill/releases/tag/v3.0.0

@mydea - Have time to send a PR here to update to the new API?

mydea:
Sure thing!

rwjblue:
OK, I think this is resolved now. Closing (happy to reopen if I've missed something).

PR #357, opened 2020-07-29
Update babel-plugin-ember-modules-api-polyfill from 2.13.4 to 3.0.0

mydea:
Migrate the term `blacklist` to `ignore` for that dependency.

See https://github.com/babel/ember-cli-babel/issues/355


rwjblue:
Restarting CI...

PR #367, opened 2020-11-18
[Chore] Move the generation of babel options into a separate function

suchitadoshi1987:
The motivation behind this change is to eventually leverage the `getBabelOptions` module to be ingested by the `babelrc` config preset.


suchitadoshi1987:
cc @rwjblue

rwjblue (commented):
Looking good, thanks for picking this up! 

I think there are still some more of the helper methods that were copied into the external module that need to be deleted from index.js.

suchitadoshi1987:
I have cleaned up a bunch of methods that are not needed in `index.js`, however, I have left a few in there since we are using them in the tests directly. Until we figure out a path forward to update the tests, I am leaving them intact. Happy to sync on the same.

rwjblue (commented):
Ya, we should move the tests to a separate file (get-babel-options-test.js or something like that) and then export the additional functions from the new file.

lib/babel-options-util.js
```diff
@@ -0,0 +1,476 @@
...
+const clone = require("clone");
+const semver = require("semver");
+
+const APP_BABEL_RUNTIME_VERSION = new WeakMap();
+
+const findApp = require("./find-app");
+const defaultShouldIncludeHelpers = require("./default-should-include-helpers");
+
+module.exports = {
+  _getPresetEnv(config, project) {
```
rwjblue, line 11: Let's roll this back to an earlier format. We do need to export, but I'd like to be able to call the functions _without_ `this`.

Something like:

```js
function getPresetEnv(config, project) {
}

//...snip...

module.exports = {
  _getPresetEnv: getPresetEnv,
}
```

lib/get-babel-options.js
```diff
@@ -0,0 +1,79 @@
+const babelOptionsUtil = require('./babel-options-util');
+
+module.exports = function getBabelOptions(config, appInstance) {
+  let { parent, project } = appInstance;
+  let addonProvidedConfig = babelOptionsUtil._getAddonProvidedConfig(config);
+  let shouldCompileModules = babelOptionsUtil._shouldCompileModules(config, project);
+  let shouldIncludeHelpers = babelOptionsUtil._shouldIncludeHelpers(config, appInstance);
+  let shouldHandleTypeScript = babelOptionsUtil._shouldHandleTypeScript(config, parent);
+  let shouldIncludeDecoratorPlugins = babelOptionsUtil._shouldIncludeDecoratorPlugins(config);
```
rwjblue, line 9: I'd like to destructure these up above, so we are just calling functions directly (basically the same comment as my other one RE: relying on `this`)

lib/get-babel-options.js
```diff
@@ -0,0 +1,79 @@
...
+    .concat(
+      shouldIncludeHelpers && babelOptionsUtil._getHelpersPlugin(project),
+      userPlugins,
+      babelOptionsUtil._getDebugMacroPlugins(config),
+      babelOptionsUtil._getEmberModulesAPIPolyfill(config, parent, project),
+      babelOptionsUtil._getEmberDataPackagesPolyfill(config, parent),
+      shouldCompileModules && babelOptionsUtil._getModulesPlugin(),
+      userPostTransformPlugins
+      )
+      .filter(Boolean);
```
rwjblue, line 64: Indentation here seems somewhat odd, is it right?

rwjblue (commented):
This is really great progress!

lib/babel-options-util.js
suchitadoshi1987: sure thing, if we do that then we will have to export all the helper methods, which I am not sure if its helpful. May be I could just export the functions that I am leveraging in the tests rather than all of them

rwjblue (approved):

PR #368, opened 2020-11-18
fix: avoid assuming host app has `ember-cli-babel` installed

alexlafroscia:
The original code here assumes that searching for `ember-cli-babel` in the host app's addons always has a "hit". However, this might not be the case, if the host app is trying to experiment with some other transpiler.

The change here allows the for lookup of the package version in the host app's dependencies to return `undefined` without an error related to accessing `pkg` on `undefined`.

### DIFF
--- a/lib/babel-options-util.js
+++ b/lib/babel-options-util.js
@@ -446,11 +446,11 @@ function _shouldIncludeHelpers(options, appInstance) {
     shouldIncludeHelpers = defaultShouldIncludeHelpers(project);
   }
 
-  let appEmberCliBabelPackage = project.addons.find(
+  let appEmberCliBabel = project.addons.find(
     (a) => a.name === "ember-cli-babel"
-  ).pkg;
+  );
   let appEmberCliBabelVersion =
-    appEmberCliBabelPackage && appEmberCliBabelPackage.version;
+    appEmberCliBabel && appEmberCliBabel.pkg && appEmberCliBabel.pkg.version;
 
   if (
     appEmberCliBabelVersion &&
--- a/node-tests/addon-test.js
+++ b/node-tests/addon-test.js
@@ -900,6 +900,12 @@ describe('ember-cli-babel', function() {
       expect(_shouldIncludeHelpers({}, this.addon)).to.be.false;
     });
 
+    it('should work when the host app does not include ember-cli-babel', function() {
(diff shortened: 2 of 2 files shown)


rwjblue (commented):
Definitely seems good to me, would you mind adding a test to make sure we don't regress again?

alexlafroscia:
Happy to! Could you point me in a direction toward how I could do that?

My guess would be some kind of new test case in this file, that configures an faux addon that has `ember-cli-babel` as a dependency but an app that does not. Does that sound right? 

https://github.com/babel/ember-cli-babel/blob/master/node-tests/addon-test.js

rwjblue:
Yes, exactly. 


You'd make a test for the function in this section:

https://github.com/babel/ember-cli-babel/blob/036723709e26aba68eee1d5a9df49f97944e435b/node-tests/addon-test.js#L922

Similar to this test:

https://github.com/babel/ember-cli-babel/blob/036723709e26aba68eee1d5a9df49f97944e435b/node-tests/addon-test.js#L929-L931

But just before the expectation, set `project.addons` to an empty array. Normally, the `ember-cli-babel` instance is pushed into `project.addons` here:

https://github.com/babel/ember-cli-babel/blob/036723709e26aba68eee1d5a9df49f97944e435b/node-tests/addon-test.js#L57

alexlafroscia:
Awesome! I'll take care of that soon. Thanks for some help planning out how to test this!

rwjblue:
Just a heads up, I just merged https://github.com/babel/ember-cli-babel/pull/367 which moves the files that I linked to above around a little bit. It shouldn't change any of the specifics though.

alexlafroscia:
I went and rebased from `master` to make sure this still works with the PR @rwjblue mentioned above merged.

I made sure the test I added actually fails if I don't include the "fix" that is included as well. It fails without it and passes with it!

rwjblue:
Dang, looks like we have another conflict. Sorry for letting this linger @alexlafroscia... 😞

rwjblue:
Did the rebase locally...

rwjblue (approved):

alexlafroscia:
Sorry I didn't get to this before you did. Thanks for re-basing this PR for me and getting it merged @rwjblue! 🙌

PR #371, opened 2020-11-30
Adds support for babel config file

suchitadoshi1987:
- [x] cleanup  index.js  - it uses the babel-utils functions and remove the duplicate functionalities
- [x] default babel options are now the responsibility of the transpileTree
- [x] cleaned up the code from getBabelOptions  file where it should just return the plugin and preset config
- [x] exported the helper methods from emberPlugins
- [x] fix existing tests
- [x] add tests for `babelrc` config
- [x] add README for usage of babelConfig


.ember-cli
```diff
@@ -1,9 +1,3 @@
 {
-  /**
-    Ember CLI sends analytics information by default. The data is completely
```
suchitadoshi1987, line 3: need to revert this change

.babelrc.js
```diff
@@ -0,0 +1,50 @@
...
+      [
+        require.resolve("@babel/plugin-transform-runtime"),
+        {
+          version: require("@babel/plugin-transform-runtime/package").version,
+          regenerator: false,
+          useESModules: true,
+        },
+      ],
+      [require.resolve("@babel/plugin-proposal-decorators"), { legacy: true }],
+      [require.resolve("@babel/plugin-proposal-class-properties")],
```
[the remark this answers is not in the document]
suchitadoshi1987: yep need to pull those into the `ember-plugins`

README.md
```diff
@@ -291,7 +292,77 @@ module.exports = function(defaults) {
...
+};
+```
+
+#### Ember Plugins
+
+Ember Plugins is a helper function that returns a list of plugins that are required for transpiling Ember correctly. You can import this helper function and add it to your existing `babel.config` file.
+
+**Config options:**
+
+```json
```
rwjblue, line 357: Lets swap this to `js` file type so it highlights better in the diff

README.md
```diff
@@ -291,7 +292,77 @@ module.exports = function(defaults) {
...
+      [
+        require.resolve("@babel/plugin-transform-runtime"),
+        {
+          version: require("@babel/plugin-transform-runtime/package").version,
+          regenerator: false,
+          useESModules: true,
+        },
+      ],
+      [require.resolve("@babel/plugin-proposal-decorators"), { legacy: true }],
+      [require.resolve("@babel/plugin-proposal-class-properties")],
```
rwjblue, line 342: I think we need to move this into `emberPlugins` isntead of requiring it first

README.md
```diff
@@ -291,7 +292,77 @@ module.exports = function(defaults) {
...
+
+      // this is where all the ember required plugins would reside
+      ...emberPlugins({ ...options }),
```
rwjblue, line 345: Lets tweak this to passing `__dirname`

rwjblue (changes requested):

README.md
suchitadoshi1987: 👍

index.js
```diff
@@ -35,31 +55,85 @@ module.exports = {
...
     let postDebugTree = this._debugTree(inputTree, `${description}:input`);
-
-    let options = this.buildBabelOptions(config);
+    let options = this._getDefaultBabelOptions(config);
     let output;
-    if (this._shouldDoNothing(options)) {
+
+    const babelConfigFile = ROOT_CONFIG_FILENAMES.find((fileName) =>
+      fs.existsSync(path.resolve(this.parent.root, fileName))
+    );
```
rwjblue, line 113: We should avoid this I/O if we haven't been opted in to.

index.js
```diff
@@ -35,31 +55,85 @@ module.exports = {
...
+
+    const customAddonConfig = config['ember-cli-babel'];
+    const shouldUseBabelConfigFile = customAddonConfig && customAddonConfig.useBabelConfig && babelConfigFile;
+
+    if (shouldUseBabelConfigFile) {
+      options = Object.assign({}, options, {
+        "presets":[ require.resolve(path.resolve(this.parent.root, babelConfigFile)) ],
```
rwjblue, line 120: Isn't babelConfigFile here already resolved?

lib/ember-plugins.js
```diff
@@ -0,0 +1,167 @@
...
+function _getProposalDecoratorsAndClassPlugins(config) {
+  if (!config.shouldIgnoreDecoratorAndClassPlugins) {
+    return [
+      ["@babel/plugin-proposal-decorators", { legacy: true }],
+      ["@babel/plugin-proposal-class-properties"],
+    ];
+  }
+}
+
+module.exports = function (appRoot, config = {}) {
```
rwjblue, line 151: Can you add some API docs for this file that list the possible options that are allowed (as it is, it's kinda hard to see the full list of possible options)?

README.md
```diff
@@ -291,7 +292,75 @@ module.exports = function(defaults) {
...
+      [
+        require.resolve("@babel/plugin-transform-runtime"),
+        {
+          version: require("@babel/plugin-transform-runtime/package").version,
+          regenerator: false,
+          useESModules: true,
+        },
+      ],
+      // this is where all the ember required plugins would reside
+      ...emberPlugins(__dirname, { ...options }),
```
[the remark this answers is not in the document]
suchitadoshi1987: not really, was just a depiction that users can add in the config options if they'd like to

README.md
```diff
@@ -291,7 +292,75 @@ module.exports = function(defaults) {
...
+    useBabelConfig: true,
+    // ember-cli-babel related options
+  },
+});
+```
+
+```js
+//babel.config.js
+
+const { emberPlugins } = require("ember-cli-babel");
```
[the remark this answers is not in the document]
suchitadoshi1987: good catch!

index.js
suchitadoshi1987: no, the `babelConfigFile` is only the file name that was found based on the path. I ll change that part so that it returns the whole path instead of just the file name.

index.js
```diff
@@ -35,31 +46,101 @@ module.exports = {
...
+      if (!babelConfigFile) {
+        throw new Error(
+          "Missing babel config file in the project root. Please double check if the babel config file exists or turn off the `useBabelConfig` option in your ember-cli-build.js file."
+        );
+      }
+      options = Object.assign({}, options, {
+        presets: [
+          require.resolve(babelConfigFile),
+        ],
+      });
```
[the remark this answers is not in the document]
suchitadoshi1987: absolutely! for some unexplainable vague reason, the time I was developing this, I read in the docs to add in the file in presets 😑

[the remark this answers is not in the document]
suchitadoshi1987: `loadPartialConfig` returns an object of [ConfigItems](https://babeljs.io/docs/en/babel-core/#configitem-type) which is not liked by broc transpiler since it expects a POJO. Since we are doing a hybrid approach where we are injecting our own options along with the config, one way we could solve this is by reading the config file inline and then pass it to the options.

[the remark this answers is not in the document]
rwjblue: We chatted through this a bit in DM / calls, we are going to update to use `configFile` path from `.loadPartialConfig` which will properly be serializable and leave all of the identification and invocation of the config file up to babel itself.

Issue #372, opened 2020-12-02
Renamed imports are not propragated to call site


NullVoxPopuli:
```ts
import { assert as debugAssert } from '@ember/debug';

debugAssert('', condition);
// becomes
Ember.debugAssert('', condition);
// instead of
Ember.assert('', condition);
```

rwjblue:
Can you open up a failing test case in `babel-plugin-ember-modules-api-polyfill`?

rwjblue:
Should be straightforward to fix over there though, thanks for working on it!

PR #373, opened 2020-12-02
Ember.assert is not renamed when import { assert as X } is used

NullVoxPopuli:
Ref: https://github.com/babel/ember-cli-babel/issues/372

### DIFF
--- a/node-tests/addon-test.js
+++ b/node-tests/addon-test.js
@@ -153,7 +153,8 @@ describe('ember-cli-babel', function() {
       it("should replace imports by default", co.wrap(function* () {
         input.write({
           "foo.js": `import Component from '@ember/component'; Component.extend()`,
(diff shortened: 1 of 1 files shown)


ro0gr:
just noticed the same issue for `deprecate`:
```js
import { deprecate as emberDeprecate } from '@ember/application/deprecations';

// ....
deprecate(message, // Here it fails with "TypeError: Ember.emberDeprecate is not a function"
```

worked it around by avoiding named import alias https://github.com/san650/ember-cli-page-object/pull/537/commits/2d7ea1b22f3ac910381c2f0b009fc6d9db293691

rwjblue:
Haven't had a chance to dig in here. I suspect the issue is in babel-plugin-debug-macros though.

NullVoxPopuli:
this is no longer an issue with latest ember-cli-babel (8+, did not test earlier)

PR #382, opened 2021-02-22
Use useEmberModule from modules plugin to deprecate global Ember

NullVoxPopuli:
Related: 
 - emberjs/ember.js#19357
 - https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill/pull/175

### DIFF
--- a/lib/ember-plugins.js
+++ b/lib/ember-plugins.js
@@ -91,7 +91,7 @@ function _shouldDeprecateGlobalEmber(appRoot) {
   }
 
   let pkg = require(packagePath);
-  return pkg && semver.gte(pkg.version, "3.26.0-beta.1");
+  return pkg && semver.gte(pkg.version, "3.27.0-alpha.1");
 }
 
 function _emberDataVersionRequiresPackagesPolyfill(appRoot) {
--- a/lib/ember-plugins.js
+++ b/lib/ember-plugins.js
@@ -60,9 +60,13 @@ function _getEmberModulesAPIPolyfill(appRoot, config) {
 
   if (_emberVersionRequiresModulesAPIPolyfill()) {
     const ignore = _getEmberModulesAPIIgnore(appRoot, config);
+    const useEmberModule = _shouldDeprecateGlobalEmber(appRoot);
 
     return [
-      [require.resolve("babel-plugin-ember-modules-api-polyfill"), { ignore }],
+      [
+        require.resolve("babel-plugin-ember-modules-api-polyfill"),
+        { ignore, useEmberModule },
+      ],
     ];
   }
 }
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
I don't know how to test this, since we are `require`ing packages in the ember-plugins file

lib/ember-plugins.js
```diff
@@ -80,6 +84,16 @@ function _shouldIgnoreJQuery(appRoot) {
...
 }
 
+function _shouldDeprecateGlobalEmber(appRoot) {
+  let packagePath = resolvePackagePath("ember-source", appRoot);
+  if (packagePath === null) {
+    return false;
+  }
+
+  let pkg = require(packagePath);
+  return pkg && semver.gte(pkg.version, "3.26.0-beta.1");
```
rwjblue, line 94:
```suggestion
  return pkg && semver.gte(pkg.version, "3.27.0-alpha.1");
```

rwjblue (changes requested):
We can test with `fixturify-project`

patocallaghan:
👋 @rwjblue I spoke to @NullVoxPopuli and I offered to get this PR over the line for him. I just have a question.

I see we've already updated `ember-plugins.js` to use `useEmberModule` but do we also need to add `useEmberModule` to `babel-options-util.js` too? What's the difference between the two files/usages?

https://github.com/babel/ember-cli-babel/blob/20924faa529cf4f9c6f7cb8fa2c89d968184b620/lib/babel-options-util.js#L132

patocallaghan:
@rwjblue @NullVoxPopuli I've added a test to `node-tests/addon-test.js` which was set up already to use `fixturify-project`. It seems though these tests don't exercise the logic in `lib/ember-plugins.js` so I had to add the `useEmberModule` logic to `lib/babel-options-util.js` too.

patocallaghan:
hmm, in the `floating dependencies` CI tests, i.e. `yarn install --no-lockfile`, the ember module gets transpiled to `_ember2` rather than `_ember`. I'm able to reproduce this locally. 

image

Any idea how the name of the AMD module is determined? why would it appear as `_ember2` in this scenario?

rwjblue:
@patocallaghan - I think we should bump the babel-plugin-* that we have as dependencies, those failures are due to an update in one of those and bumping + fixing the fixtures in another PR should unblock.

patocallaghan:
I've put together a PR in https://github.com/babel/ember-cli-babel/pull/385 which upgrades some deps. It mimics what we see with the "floating deps" in CI so it will resolve our issue

rwjblue:
Hmm, I expected that to fail. Perhaps the issue isn't with one of those plugins?

patocallaghan:
Ah no it is. Once I rebase this PR against that branch it starts failing. I've updated locally with _ember2 and it works fine

patocallaghan:
I should say nothing else fails though

patocallaghan:
@rwjblue I've rebased this against master to pick up the package upgrades. CI looks happy now 🎉

pzuraq (approved):

patocallaghan:
@rwjblue any chance this could be merged and released please? I need this and the package upgrade PR in some other repo PRs

rwjblue:
Closing in favor of #387

Issue #384, opened 2021-03-01
[v7.21.0] includeExternalHelpers option breaking with latest @babel


hoIIer:
I cleared node_modules and ran `npm install` with version `7.21.0` and when I run `ember serve` I get the following error:

```
Error: Could not find module @babel/runtime/helpers/initializerDefineProperty/_index.mjs imported from @babel/runtime/helpers/esm/initializerDefineProperty
    at missingModule (/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/broccoli-276094EuqwRsv8Chf/out-559-broccoli_merge_trees/assets/vendor/loader/loader.js:247:1)
    at findModule (/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/broccoli-276094EuqwRsv8Chf/out-559-broccoli_merge_trees/assets/vendor/loader/loader.js:258:1)
    at Module.findDeps (/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/broccoli-276094EuqwRsv8Chf/out-559-broccoli_merge_trees/assets/vendor/loader/loader.js:168:1)
    at findModule (/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/broccoli-276094EuqwRsv8Chf/out-559-broccoli_merge_trees/assets/vendor/loader/loader.js:262:1)
    at Module.findDeps (/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/broccoli-276094EuqwRsv8Chf/out-559-broccoli_merge_trees/assets/vendor/loader/loader.js:168:1)
    at findModule (/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/broccoli-276094EuqwRsv8Chf/out-559-broccoli_merge_trees/assets/vendor/loader/loader.js:262:1)
    at requireModule (/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/broccoli-276094EuqwRsv8Chf/out-559-broccoli_merge_trees/assets/vendor/loader/loader.js:24:1)
    at Class._extractDefaultExport (/var/folders/9r/gp7xyl_j0l3flcwp78j7mjcr0000gn/T/broccoli-276094EuqwRsv8Chf/out-559-broccoli_merge_trees/assets/addon-tree-output/ember-resolver/resolvers/classic/index.js:445:1)
[...]

PR #385, opened 2021-03-07
Upgrade babel-plugins

patocallaghan:
Upgrading all `babel-plugin-*` dependencies due to conversation in https://github.com/babel/ember-cli-babel/pull/382#issuecomment-792146538

/cc @rwjblue


rwjblue (approved):

rwjblue (commented):
Can you also update `package.json` to use those new minimum versions as well?

patocallaghan:
Updated the package.json 👍

rwjblue:
Hmmm. Something is causing the floating dependencies tests to time out.

patocallaghan:
@rwjblue seeing this in CI. Will have a dig later

image

rwjblue:
Seems like the same thing being reported in later comments over in https://github.com/babel/ember-cli-babel/issues/384

patocallaghan:
Interesting, so bumping a few other babel-related deps has caused the issue to manifest in the non-"floating dependencies" tests. Unfortunately it's only happening in CI for me, I can't reproduce locally, so will need to rollback each of those upgrades I just did one-by-one.

image

rwjblue:
I've canceled some of the CI runs (to try to get faster turn around)

patocallaghan:
@rwjblue it's not pretty but I've pinned `@babel/runtime` to 7.12.18 and that resolves the floating dependencies issue. I haven't bottomed-out out on it yet but it appears to be related to https://github.com/babel/babel/issues/12881. This issue was marked as resolved but bumping to the latest babel/runtime caused it to still happen.

Unfortunately as I stated earlier, I can only reproduce the issue on CI which makes debugging a bit painful.

boris-petrov:
FWIW I also had to pin `@babel/runtime` in my app to `7.12.18` but only because of an "old" `targets.js`. Using:

```js
[
  'chrome >= 85',
  'edge >= 85',
  'firefox >= 80',
  'safari >= 14',
]
```

Works fine with `@babel/runtime` `7.13.9`+ but using `['firefox 35']` for `targets` leads to the error and I have to use `7.12.18` because of that. Just to mention it, not sure if it's relevant here.

patocallaghan:
Thanks @boris-petrov, that could well be the case here as we add ie11 in CI.

https://github.com/babel/ember-cli-babel/blob/20924faa529cf4f9c6f7cb8fa2c89d968184b620/tests/dummy/config/targets.js#L3-L14

I guess we should keep it pinned though as officially the Ember framework still supports IE11 (until 4.0 anyway)?

PR #387, opened 2021-03-16
Do not transform imports to use the Ember global for ember-source@3.27 and higher

patocallaghan:
PR supersedes #382 

So after some discussion in [#dev-ember-js](https://discord.com/channels/480462759797063690/485447409296736276/821417976061886524) there's been a change of plan with regards deprecating Ember Global support.

Now, rather than using `babel-plugin-ember-modules-api-polyfill`'s `useEmberModule` option to transform packages to their global equivalents and use an imported `Ember`, the plan is to no longer transform Ember package imports to use the Ember Global.

#### Source file
```js
import Component from '@ember/component'; Component.extend();
```

#### Versions of `ember-source` less than 3.27.0-alpha.1
```js
define("foo", [], function () {\n "use strict";\n\n Ember.Component.extend();\n});
```

#### Versions of `ember-source` greater than or equal to 3.27.0-alpha.1
```js
define("foo", ["@ember/component"], function (_component) {\n "use strict";\n\n _component.default.extend();\n});
```

### DIFF
--- a/lib/ember-plugins.js
+++ b/lib/ember-plugins.js
@@ -46,11 +46,14 @@ function _getDebugMacroPlugins() {
     ],
   ];
 }
-function _emberVersionRequiresModulesAPIPolyfill() {
-  // once a version of Ember ships with the
-  // emberjs/rfcs#176 modules natively this will
-  // be updated to detect that and return false
-  return true;
+function _emberVersionRequiresModulesAPIPolyfill(appRoot) {
+  let packagePath = resolvePackagePath("ember-source", appRoot);
+  if (packagePath === null) {
+    return true;
+  }
+
+  let pkg = require(packagePath);
+  return pkg && semver.lt(pkg.version, "3.27.0-alpha.1");
 }
 
 function _getEmberModulesAPIPolyfill(appRoot, config) {
@@ -58,7 +61,7 @@ function _getEmberModulesAPIPolyfill(appRoot, config) {
     return;
   }
 
-  if (_emberVersionRequiresModulesAPIPolyfill()) {
+  if (_emberVersionRequiresModulesAPIPolyfill(appRoot)) {
     const ignore = _getEmberModulesAPIIgnore(appRoot, config);
 
     return [
@@ -128,7 +131,7 @@ function _getEmberDataPackagesPolyfill(appRoot, config) {
 function _getModuleResolutionPlugins(config) {
   if (!config.disableModuleResolution) {
     const resolvePath = require("../lib/relative-module-paths")
-    .resolveRelativeModulePath;
+      .resolveRelativeModulePath;
     return [
       [require.resolve("babel-plugin-module-resolver"), { resolvePath }],
       [
--- a/node-tests/addon-test.js
+++ b/node-tests/addon-test.js
@@ -256,12 +256,12 @@ describe('ember-cli-babel', function() {
       it(
         "should replace imports with Ember Globals",
         co.wrap(function* () {
-          const PRE_GLOBAL_RESOLVER_DEPRECATION_VERSION = "3.26.0";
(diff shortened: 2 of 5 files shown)


lib/ember-plugins.js
```diff
@@ -46,19 +46,22 @@ function _getDebugMacroPlugins() {
     ],
   ];
 }
-function _emberVersionRequiresModulesAPIPolyfill() {
-  // once a version of Ember ships with the
-  // emberjs/rfcs#176 modules natively this will
-  // be updated to detect that and return false
-  return true;
+function _emberVersionRequiresModulesAPIPolyfill(appRoot) {
```
rwjblue, line 49: Can we use this version in both cases (essentially replacing the changes in `lib/babel-options-util.js`)?

node-tests/addon-test.js
```diff
@@ -216,6 +216,112 @@ describe('ember-cli-babel', function() {
...
+      it(
+        "should replace imports with Ember Globals",
+        co.wrap(function* () {
+          const PRE_GLOBAL_RESOLVER_DEPRECATION_VERSION = "3.26.0";
+          dependencies[
+            "ember-source"
+          ] = PRE_GLOBAL_RESOLVER_DEPRECATION_VERSION;
+          input.write(
+            buildEmberSourceFixture(PRE_GLOBAL_RESOLVER_DEPRECATION_VERSION)
+          );
```
rwjblue, line 265: Why is this variable named `PRE_GLOBAL_RESOLVER_DEPRECATION_VERSION`?

rwjblue: AHA! I think you mean "global resolver" as in transforming to `Ember.Foo`. Unfortunately, we actually already have a thing that is called "globals resolver" and it is in fact deprecated too. 

Sorry for the annoyance, but lets find a different name that doesn't cause future rwjblue confusion.

lib/ember-plugins.js
patocallaghan: So just to clarify, you're saying that we should use the `resolvePackagePath` way of `ember-source` detection in the `babel-options-util.js`, rather than use `VersionChecker`?

If that is correct, I tried that in an earlier iteration but `resolvePackagePath` throws the following error when running `yarn test`:

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type object
      at validateString (internal/validators.js:112:11)
      at Object.resolve (path.js:980:7)
      at resolvePackagePath (node_modules/resolve-package-path/lib/resolve-package-path.js:221:30)
      at resolvePackagePath (node_modules/resolve-package-path/index.js:86:19)
      at _emberVersionRequiresModulesAPIPolyfill (lib/babel-options-util.js:108:21)
      at _getEmberModulesAPIPolyfill (lib/babel-options-util.js:137:7)
      at getBabelOptions (lib/get-babel-options.js:55:7)
      at Class.buildBabelOptions (index.js:49:12)
      at Class.transpileTree (index.js:131:49)
      at Context.<anonymous> (node-tests/addon-test.js:1848:26)
      at Generator.next (<anonymous>)
      at onFulfilled (node_modules/co/index.js:65:19)
```

I then noticed we use `VersionChecker` approach throughout `babel-options-util.js` so went with that.

node-tests/addon-test.js
patocallaghan: Does `PRE_EMBER_MODULE_IMPORTS_VERSION` sound better to you?

rwjblue: Ya, sounds great. Apologies again for the bike shedding.

lib/ember-plugins.js
rwjblue: In the case that failed were you passing a string in? Seems odd...

pzuraq (approved):

rwjblue (approved):

Issue #389, opened 2021-03-18
7.24.0+ Breaks everything


boris-petrov:
Upgrading to 7.24.0 or 7.25.0 leads to a bunch of errors like:

```
Error: Could not find module `lodash/identity` imported from `...`
```

Everything else is the latest possible version. Do you need a reproduction?

makepanic:
Do you use ember-cli-typescript? It might be related to https://github.com/ef4/ember-auto-import/issues/359

boris-petrov:
I do use it, yes. It's possible, I guess. Thanks, I'll subscribe to the other issue too!

wagenet:
Also seeing this.

rwjblue:
Fix released in 7.26.1

wagenet:
Thanks!

Issue #386, opened 2021-03-08
Upgrading from 7.23.1 to 7.24.0 breaks various NPM module imports


kpfefferle:
I am unable to upgrade some of my projects to ember-cli-babel@7.24.0 because it breaks importing from some NPM modules. So far I have seen this issue with the following packages:

- [Faker](https://github.com/marak/Faker.js/): `import faker from 'faker';`
- [Luxon](https://moment.github.io/luxon/): `import { DateTime } from 'luxon';`

Both of these worked fine in my Ember projects up through ember-cli-babel@7.23.1, but with ember-cli-babel@7.24.0 they error out with `Error: Could not find module [MODULE] imported from [FILE]`

rwjblue:
Can you make a small repro?

kpfefferle:
My initial attempt to make a [small repro](https://github.com/kpfefferle/ember-cli-babel-package-imports) using only these addons is not generating the same errors I was seeing in my other application. I'll close this for now while I investigate what else might be causing this, and I'll reopen later if needed.

rwjblue:
kk, thank you for reporting back! Please do let me know if you are able to track down a repro, I definitely want to make sure we aren't breaking things for folks...

PR #392, opened 2021-03-18
Fix compatibility with ember-auto-import.

rwjblue:
ember-cli-babel@7.24 accidentally broke compatibility with ember-auto-import when it fixed `buildBabelOptions` to only return custom babel config **not** broccoli-babel-transpiler config.

This fixes that compatiblity in the case where `babelAddon.buildBabelOptions()` is called, but still preserves the ability to get _either_ the broccoli-babel-transpiler config _or_ the `@babel/core` config.

Fixes https://github.com/ef4/ember-auto-import/issues/359
Fixes https://github.com/babel/ember-cli-babel/issues/389
Related to https://github.com/ef4/ember-auto-import/pull/362

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,7 +380,7 @@ interface EmberCLIBabel {
   /**
     Used to generate the options that will ultimately be passed to babel itself.
   */
-  buildBabelOptions(config?: EmberCLIBabelConfig): Opaque;
+  buildBabelOptions(type: 'babel' | 'broccoli', config?: EmberCLIBabelConfig): Opaque;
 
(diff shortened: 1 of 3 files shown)


kpfefferle:
Thanks @rwjblue, this seems to have fixed the projects I had that were failing https://github.com/babel/ember-cli-babel/issues/386! 🎉

rwjblue:
Thanks for confirming @kpfefferle! I didn't realize this was also the cause there, but it definitely makes sense.

PR #394, opened 2021-03-19
Include addon name directory in 'treeForAddon'

charlespierce:
Info
-----
* The default `treeForAddon` method provided by `ember-cli` ensures that all compiled assets are put into a directory named after the addon itself (see https://github.com/ember-cli/ember-cli/blob/9911f5493c00986ac2d0be104a80152426240a30/lib/models/addon.js#L1196)
* With its custom `treeForAddon` method, however, `ember-cli-babel` currently doesn't do that for the babel helpers.
* This can cause issues when the build system (e.g. [Embroider](https://github.com/embroider-build/embroider)) expects that behavior from an addon.

Changes
-----
* Added a `Funnel` that includes the addon module name directory when building the Babel helpers (using the same `moduleName` method that the base addon class uses.

### DIFF
--- a/index.js
+++ b/index.js
@@ -245,11 +245,15 @@ module.exports = {
       destDir: '@babel/runtime/helpers/esm'
     });
 
-    return this.transpileTree(babelHelpersTree, {
+    const transpiledHelpers = this.transpileTree(babelHelpersTree, {
       'ember-cli-babel': {
         // prevents the helpers from being double transpiled, and including themselves
         disablePresetEnv: true
       }
+    });
+
+    return new Funnel(transpiledHelpers, {
+      destDir: this.moduleName(),
     });
   },
 


index.js
```diff
@@ -245,12 +245,16 @@ module.exports = {
...
+    const transpiledHelpers = this.transpileTree(babelHelpersTree, {
       'ember-cli-babel': {
         // prevents the helpers from being double transpiled, and including themselves
         disablePresetEnv: true
       }
     });
+
+    return new Funnel(transpiledHelpers, {
+      destDir: this.moduleName(),
+    });
```
rwjblue, line 257: Doesn't this mean that the modules aren't in the right place when they are imported? Or maybe the answer is: it doesn't matter because the define statements (created by `this.transpileTree` just above) will the right module names (unscoped to `ember-cli-babel`).

Would you mind double checking in a `ember new` style app with and without Embroider (with the flag turned on) to confirm that things work properly (and the modules end up in the build with the right module paths)?

rwjblue: I dug into the output and confirmed that this is indeed how it works. Sorry for the noise @charlespierce.

PR #400, opened 2021-04-23
Update `_shouldHandleTypeScript` to read from `pkg.dependencies` instead of relying on `addons`

brendenpalmer:
This fixes a bug between `ember-cli-typescript`, `ember-engines` dedupe, and `ember-cli-babel`. `ember-cli-babel` uses the existence of `ember-cli-typescript` as a dependency to determine whether to process `*.ts` files; however, `ember-engines` dedupe is _stateful_ so it's possible for `ember-cli-typescript` to not be a dependency (part of `addons`) when `ember-cli-babel` is running.

For more info on `ember-engines` dedupe logic: https://github.com/ember-engines/ember-engines/blob/master/packages/ember-engines/lib/utils/deeply-non-duplicated-addon.js#L35

### DIFF
--- a/lib/babel-options-util.js
+++ b/lib/babel-options-util.js
@@ -407,6 +407,26 @@ function _shouldIncludeDecoratorPlugins(config) {
   return customOptions.disableDecoratorTransforms !== true;
 }
 
+/**
+ * Returns whether we should handle TypeScript (based on the existence of
+ * `ember-cli-typescript` as a depenency). It's worth noting that we parse
+ * the `package.json` deps/devDeps directly (rather than using `addons` on
+ * the parent) because it's possible for `ember-cli-typescript` not to exist
+ * on the addons array, even if it is a dependency.
+ *
+ * Some more context:
+ *
+ * `ember-cli-typescript` returns a stable cache key so its possible for it to
+ * be deduped as part of `ember-engines`. The reason this is important is because
+ * `ember-engines` dedupe is _stateful_ so it's possible for `ember-cli-typescript`
+ * to not be part of the addons array when `ember-cli-babel` is running.
+ * 
+ * For more info on `ember-engines` dedupe logic:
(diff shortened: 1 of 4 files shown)


lib/babel-options-util.js
```diff
@@ -406,16 +407,23 @@ function _shouldIncludeDecoratorPlugins(config) {
...
-    parent.addons.find((a) => a.name === "ember-cli-typescript");
-  return (
-    typeof typeScriptAddon !== "undefined" &&
-    semver.gte(typeScriptAddon.pkg.version, "4.0.0-alpha.1")
-  );
+
+  let tsDependency =
+    parent.pkg &&
+    parent.pkg.dependencies &&
+    parent.pkg.dependencies["ember-cli-typescript"];
```
rwjblue, line 418: I think we need to consider what to do when `this.parent === this.project`, in that case we need to look at `devDependencies` as well.

lib/babel-options-util.js
```diff
@@ -406,16 +407,23 @@ function _shouldIncludeDecoratorPlugins(config) {
...
+
+  if (typeof tsDependency !== "undefined") {
```
rwjblue, line 420: We can probably just `tsDependency !== undefined` (no need for typeof check here)

lib/babel-options-util.js
```diff
@@ -406,16 +407,23 @@ function _shouldIncludeDecoratorPlugins(config) {
...
+    let tsPkgPath = resolvePackagePath("ember-cli-typescript", parent.root);
+    let tsPkg = tsPkgPath ? require(tsPkgPath) : null;
+    return tsPkg ? semver.gte(tsPkg.version, "4.0.0-alpha.1") : false;
```
rwjblue, line 423:
```suggestion
    if (tsPkgPath === null) {
      return false;
    }
    let tsPkg = require(tsPkgPath);
    return semver.gte(tsPkg.version, "4.0.0-alpha.1");
```

rwjblue (changes requested):

rwjblue (approved):

PR #409, opened 2021-06-06
Update workerpool

mansona:
This is an experiment to try and update workerpool. See https://github.com/babel/broccoli-babel-transpiler/pull/203 for more information.


rwjblue:
Hmm, looks like this contains commits that have already landed on master. Can you rebase to drop those out?

mansona:
@rwjblue actually I just had the wrong base. This would need to be a v6 patch release but depends on https://github.com/babel/broccoli-babel-transpiler/pull/203 also being patch released too 🙈 I was planning to put together a better explanation for you on this but it has been a mad week

bertdeblock:
Closing as `broccoli-babel-transpiler` now includes `workerpool` v6, so it seems this is no longer needed.

Issue #425, opened 2021-12-16
7.26.7: Decorating class property failed. Please ensure that proposal-class-properties is enabled and runs after the decorators transform.


NullVoxPopuli:
Just got this in an OSS project (which I'm not sharing the link to, cause I have a toooon in flux :sweat_smile: ).

but anyway, just tracking the error here for discussion. if I come to a resolution, I'll post about it.

What's interesting about this error is that it's a runtime error. The build succeeded.

Using embroider:
```
Uncaught (in promise) Error: Decorating class property failed. Please ensure that proposal-class-properties is enabled and runs after the decorators transform.
    _initializerWarningHelper chunk.a5b0aa63f7c9da1d786b.js:44777
    _class chunk.a5b0aa63f7c9da1d786b.js:40318
    PageTitle chunk.a5b0aa63f7c9da1d786b.js:40325
    Ember 4
    getHelper manager.js:863
    <anonymous> runtime.js:2042
    evaluate runtime.js:1284
    evaluateSyscall runtime.js:5064
    evaluateInner runtime.js:5020
```
    
![image](https://user-images.githubusercontent.com/199018/146299229-5a7e2781-4974-4f5a-95de-246ae5907673.png)

NullVoxPopuli:
More info -- I was able to delete addon invocations until I could work with my own code, got this: (I also turned on sourcemaps):
![image](https://user-images.githubusercontent.com/199018/146304013-4281950f-e1d0-4f41-a73d-1e48e494e846.png)

the compiled output though is this:

```js
let DemoSelect = (_class = class DemoSelect extends _glimmer_component__WEBPACK_IMPORTED_MODULE_4__["default"] {
  editor = (0,_home_nullvoxpopuli_Development_NullVoxPopuli_limber_node_modules_babel_runtime_helpers_esm_initializerWarningHelper_js__WEBPACK_IMPORTED_MODULE_1__["default"])(_descriptor, this);
  demos = _snippets__WEBPACK_IMPORTED_MODULE_8__.NAMES;

  async select(demoName) {
    let demo = await (0,_snippets__WEBPACK_IMPORTED_MODULE_8__.getFromLabel)(demoName);
    this.editor.updateDemo(demo);
  }

[...]

Issue #423, opened 2021-12-15
Class decorator woes with 7.26.7


jherdman:
Since updating to ecli-babel 7.26.7 I see the following error:

> Decorating class property failed. Please ensure that proposal-class-properties is enabled and runs after the decorators transform

This is reproduced with the following app: https://github.com/jherdman/hump-day-sadness

Edit: updated the version number 🤭

vst:
Same here. I had to pin down the dependency to v7.26.6.

PS: I think that you meant v7.26.7.

NullVoxPopuli:
Whoops, i made a duplicate https://github.com/babel/ember-cli-babel/issues/425

dknutsen:
@vst you can use a yarn resolution to force 7.26.6 temporarily

chriskrycho:
Working on a fix in #426. Will update there as we make progress. (Turns out we didn't have test coverage for this! That's the first thing I'm doing: adding coverage.)

PR #420, opened 2021-12-07
Properly handle class properties proposal

chriskrycho:
Historically, we only add the `@babel/plugin-proposal-class-properties` so that we make sure the ordering is right with the decorators proposal (otherwise, it can end up compiling in the wrong order). With a recent version of `@babel/preset-env` and, transitively, `caniuse-lite`, this resulted in cases where we added that plugin but *not* related plugins for private class properties, which in turn triggered a Babel assertion about not adding the properties together as appropriate when the caniuse database (correctly) reported that .

The fix is:

1.  Bump to a more recent version of `@babel/preset-env`, which comes with a correspondingly bumped version of `caniuse-lite`, which in turn correctly understands what the latest versions of targeted browsers are.

2.  Include in `ember-cli-babel` itself a check for whether we even *need* to add the plugin, and only provide it when the provided `targets` indicate that they require it.

Resolves #419

PR #426, opened 2021-12-16
Fix decorators babel plugin issue

chriskrycho:
Starts by introducing a failing test case for #423, #425. Will also introduce more, and hopefully will fix this. Wheeeeeee.

---

Current working theory, based on reading [the code for the Babel decorators plugin][babel-decorators]: we *must* actively, manually supply all the class properties (including private class properties and methods) plugins to make sure it works, since we're manually opting ourselves into decorators support.

If that's true, our fix will look like:

- reverting the changes we made in #420
- always adding the three class properties/fields hooks as well

[babel-decorators]: https://github.com/babel/babel/blob/0058b7fef4d028d2826645b28e3b448517cd979d/packages/babel-plugin-proposal-decorators/src/transformer-legacy.ts#L143-L157


NullVoxPopuli:
I'm curious why this "Works" then: https://github.com/NullVoxPopuli/babel-transpilation-tests
```
export default {
  presets: ['@babel/preset-env'],
  plugins: [['@babel/plugin-proposal-decorators', { legacy: true }]]
};
```
babel hurts my head lol

rwjblue:
> I'm curious why this "Works" then: NullVoxPopuli/babel-transpilation-tests

Mentioned in discord, but I'll also mention it here: that repro actually demonstrates the failure!!! 

See this assertion: https://github.com/NullVoxPopuli/babel-transpilation-tests/blob/9bbc2c8353e45bebae9ae16defd40830614718e1/output.js#L20

NullVoxPopuli:
> Mentioned in discord, but I'll also mention it here: that repro actually demonstrates the failure!!!

I'm sad to see this! lol.
I hadn't ran that code, because I figured it was small enough to "just be right" -- focusing primarily on the non-transpilation of the class properties, and the "supposed" transpilation of the decorators.
Looking at now (it's been a while since I made that repo), it's _very_ obvious that the problem is the same as what we saw with ember-cli-babel 7.26.7 :(

NullVoxPopuli:
Just tested this out, and it's an improvement!
image
I no longer see defineProperty, so this is a _huge_ improvement! thank you both <3

Issue #429, opened 2021-12-17
Babel polyfill broken in 7.26.10


nickschot:
Specifying 

```
  // ember-cli-build.js
  let app = new EmberApp(defaults, {
    'ember-cli-babel': {
      includePolyfill: true,
    },
  });
```
will cause the following error on build:
```
Build Error (SourceMapConcat)

ENOENT: no such file or directory, open '/var/folders/7h/82hn6hyn1bngg96_xvhlmq040000gn/T/broccoli-34311LQfVtgUc1OgD/out-262-broccoli_debug_debug_5_vendor_js/vendor/babel-polyfill/polyfill.js'
```

Minimal reproduction: https://github.com/nickschot/babel-polyfill-repro

Adonis320:
Affects versions 7.26.7 and upwards

rwjblue:
Hmph. Why? This is strange, we didn't change anything RE: the polypill in these versions 🤔

nickschot:
Good question 😂 I ran into it testing the updated version to see if my private class property issues were resolved.

chriskrycho:
I suspect it was an “unclean revert”[^1]: ea752c4 reverted 0d47e8b, but did *not* revert the bump to `@babel/preset-env` that [was included](https://github.com/babel/ember-cli-babel/commit/0d47e8be96b93120b0bcb261dd719231e73ef2d0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519) there. Presumably there’s a difference in what the updated `@babel/preset-env` is requiring *or not* for the polyfill.

[^1]: this sounds hilarious 😂

rwjblue:
Sorry about this y'all, fixed in https://github.com/babel/ember-cli-babel/releases/tag/v7.26.11

PR #421, opened 2021-12-13
introduce cacheKeyForTree to avoid running treeForX multiple times

RuslanZavacky:
We have a big monorepo and got to a point, where our build times are unacceptably slow. Can easily be 2-3mins to start an app. We've started to do some investigation to try and figure our what might be an issue. `cacheKeyForTree` and [this RFC](https://emberjs.github.io/rfcs/0090-addon-tree-caching.html) sparked my interest. After adding cache for all our internal addons, we've got in some cases up to 50% speed up. Next on our list is to identify all community addons that might be missing caching. `ember-cli-babel` is more/less the first one, as it might have quite a big impact.

Will use this command to show difference before an after
```
DEBUG=ember-cli:addon:cache-key-for-tree ember s &> capture-cache-key-debug.txt
```

**Log before adding a cache**
393 lines of cache opt-out. `Build successful (135792ms)`

<details>
  <summary>Show output</summary>
  
  ```
[treeFor(<internal addon, redacted> - addon-test-support) -> treeFor(<internal addon, redacted> - addon-test-support) -> treeFor(ember-cli-page-object - addon-test-support)] Opting out due to: modified methods: treeForAddonTestSupport
[treeFor(<internal addon, redacted> - addon-test-support) -> treeFor(ember-cli-page-object - addon-test-support) -> treeFor(ember-native-dom-helpers - addon-test-support)] Opting out due to: modified methods: treeForAddonTestSupport
[treeFor(@ember/test-helpers - addon-test-support)] Opting out due to: modified methods: treeForAddonTestSupport
[treeFor(<internal addon, redacted> - addon-test-support) -> treeFor(<internal addon, redacted> - addon-test-support) -> treeFor(ember-cli-page-object - addon-test-support)] Opting out due to: modified methods: treeForAddonTestSupport
[treeFor(<internal addon, redacted> - addon-test-support) -> treeFor(ember-cli-page-object - addon-test-support) -> treeFor(ember-native-dom-helpers - addon-test-support)] Opting out due to: modified methods: treeForAddonTestSupport
[treeFor(ember-sinon-qunit - addon-test-support)] Opting out due to: modified methods: treeForAddonTestSupport
[treeFor(ember-qunit - addon-test-support)] Opting out due to: modified methods: treeForAddonTestSupport
[treeFor(qunit-dom - addon-test-support)] Opting out due to: modified methods: treeForAddonTestSupport
[...]

PR #430, opened 2021-12-18
Fix missing babel polyfill in vendor tree

jaswilli:
I believe this is a fix for the mystery of the missing babel polyfill as reported in #429. It seems as though the culprit was the introduction of `cacheKeyForTree()` in #421.

Since our `treeForVendor()` conditionally adds in the babel polyfill we need to consider that condition when generating the cache key for vendor.

I verified this fixes the simple reproduction (new Ember app with the polyfill enabled in ember-cli-build) and it also fixes this same issue on a big app I have at work.

This is the first time I've gone under the covers of ember-cli-babel so I'm happy to be educated if there's a better strategy for fixing this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -275,7 +275,9 @@ module.exports = {
   },
 
   cacheKeyForTree(treeType) {
-    if (treeType === 'addon') {
+    if (treeType === 'vendor') {
+      return cacheKeyForTree('vendor', this, [this._shouldIncludePolyfill()]);
+    } else if (treeType === 'addon') {
       let isRootBabel = this.parent === this.project;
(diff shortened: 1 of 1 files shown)


Mithrilhall:
Thanks, this worked for me.

jamescdavis:
This fixes things for me as well. #421 was definitely the breaking change. @rwjblue, can we push this through?

jamescdavis (approved):

rwjblue:
Sorry for slow response here y'all, thank you @jaswilli for workin' up a fix!!

rwjblue (approved):

PR #438, opened 2022-02-23
Depend on @babel/plugin-transform-typescript v7.16.0+

### LINKED ISSUE #437: Needs bump for TS v4.5 `import type` support
Currently:

https://github.com/babel/ember-cli-babel/blob/315df8af07bdc755c7529578dd48f5e25e2930d9/package.json#L52

We need [at least v7.16.0](https://github.com/babel/babel/blob/main/CHANGELOG.md#rocket-new-feature-1) to support it. I'll open a PR shortly.

chriskrycho:
Babel v7.16.0 introduced support TypeScript 4.5's improvements to the `import type` syntax. Bump to require at least that version so ember-cli-babel users get support.

Previously, *new* users would get support by way of normal resolution for `^` dependencies, but existing codebases would end up broken if they pulled in a package which used the `import type` syntax and already had `ember-cli-babel` installed in a way which had resolved to an earlier version of the `@babel/plugin-transform-typescript` plugin.

Fixes #437.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "@babel/plugin-proposal-private-property-in-object": "^7.16.5",
     "@babel/plugin-transform-modules-amd": "^7.13.0",
     "@babel/plugin-transform-runtime": "^7.13.9",
-    "@babel/plugin-transform-typescript": "^7.13.0",
+    "@babel/plugin-transform-typescript": "^7.16.0",
     "@babel/preset-env": "^7.16.5",
     "@babel/runtime": "7.12.18",
     "amd-name-resolver": "^1.3.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -10,19 +10,53 @@
     "@jridgewell/gen-mapping" "^0.1.0"
     "@jridgewell/trace-mapping" "^0.3.9"
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.18.6":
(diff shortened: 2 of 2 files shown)


asakusuma:
@chriskrycho you may need to restart the tests. looks like there was a timeout failure.

chriskrycho:
Yeah, unfortunately I can't easily do so (don't have permissions). I might have to just re-commit and force push. 🙃

rwjblue:
Restarted build, but note this change isn't _needed_ (it just makes things a tad easier for folks to update, since they can just bump their own direct ember-cli-babel version).

mikesherov:
Hello, just to report that I've experienced the same issue as @chriskrycho here and this note doesn't remedy the situation:

> Restarted build, but note this change isn't needed (it just makes things a tad easier for folks to update, since they can just bump their own direct ember-cli-babel version).

Yarn as a dependency manager doesn't *always* pick the latest matching version of a subdependency, so while my application was directly relying upon latest `ember-cli-babel`, `ember-cli-babel` still relies on `^7.13.0` and the resulting `yarn.lock` file chose `7.13.0`, which is a *technically valid* match of the version specifier (because caret specifying earliest working major, not guaranteed latest). We need to use a yarn resolution to address this in the host app, which is not ideal.

chriskrycho:
Right, the specific reason that behaves that way is that Yarn preserves *existing* versions of transitive dependencies wherever possible, to avoid having you get broken by adding or upgrading a separate package. That's generally preferable behavior, since users *can* resolve it in a variety of ways; but it can also be annoying here.

Issue #447, opened 2022-06-08
Error: Decorating class property failed. Please ensure that proposal-class-properties is enabled and runs after the decorators transform.


NullVoxPopuli:
This error occurs when I'm consuming a V2 addon that has not complied _away_ class properties.

This is incorrect behavior, because 
 - my browsertargets all support class properties
 - the _app_ has proposal-class-properties, so it should use that
 - how does this end up working for regular v2 packages? have we always not been able to use class properties?

NullVoxPopuli:
Here is a reproduction: https://github.com/NullVoxPopuli/ember-resources/pull/616
this PR demonstrates that by not compiling away properties, you can see from a 5% to 68% reduction in files size (after minification)

the ember app gives this error:
```
Error: Decorating class property failed. Please ensure that proposal-class-properties is enabled and runs after the decorators transform.
```

NullVoxPopuli:
ah ok, getting context back now -- I saw this in the babel output:
```
function _initializerWarningHelper(descriptor, context) {
  throw new Error('Decorating class property failed. Please ensure that ' + 'proposal-class-properties is enabled and runs after the decorators transform.');
}


```
how rude

Issue #446, opened 2022-06-05
Is there a way to opt out of ember-cli-babel entirely?


NullVoxPopuli:
Related issue: https://github.com/babel/ember-cli-babel/issues/418

just wondering, given the inability to use custom babel plugins, or _not_ transpile class properties, (and a bunch of other bugs (some as old as 2019!), that seems to have been the result of ecosystem complacency in this tool in that it is "good enough", but ultimately does too much (babel-wise)).
 - https://github.com/babel/ember-cli-babel/issues/444
 - https://github.com/babel/ember-cli-babel/issues/417
 - https://github.com/babel/ember-cli-babel/issues/418
 - https://github.com/babel/ember-cli-babel/issues/296
 
 Does it even make sense to not use ember-cli-babel?
 
 _minimally_, this is all that's needed for the style of JS we write:
 ```js
 export default {
  presets: ['@babel/preset-env'],
  plugins: [
    ['@babel/plugin-proposal-decorators', { legacy: true }],
    // ... @ember/debug removal
    // ... @ember/deprecations removal
    // ... ESM to AMD
  ]
};
```
What I removed from the above:
 - class-properties -- this isn't needed because proposal-decorators doesn't need it, and browsers all natively support class properties. At least for development, removing this could speed up builds
 - ember modules api polyfill -- since ember 3.27, the modules polyfill isn't needed. Less plugins, less faster build.
 
Eventually, given that I could eventually customize ember-cli-babel, I'd like to experiment with removing the ESM to AMD transform.
 
 
 I know that addons can provide their own babel plugins / transforms, but, I don't think I'm at a point yet there addons' own configs even matter as I'm just trying to build the app and keep the `Router` in `router.js` as a native class with native properties, but today it looks like this:
 ```js
 ;define("minimal-babel-demo/router", ["exports", "@ember/routing/router", "minimal-babel-demo/config/environment"], function (_exports, _router, _environment) {
  "use strict";

  Object.defineProperty(_exports, "__esModule", {
    value: true
  });
  _exports.default = void 0;

  function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }

  class Router extends _router.default {
    constructor(...args) {
      super(...args);

      _defineProperty(this, "location", _environment.default.locationType);

      _defineProperty(this, "rootURL", _environment.default.rootURL);
    }

  }

[...]

Issue #417, opened 2021-10-14
[Bug] browserslist and targets have no effect on the output of the build (potential performance problems as well)


NullVoxPopuli:
Originally reported here: https://github.com/emberjs/ember.js/issues/19790

### 🐞 Describe the Bug
After some investigation here: https://stackoverflow.com/questions/69547969/with-babel-how-do-i-not-compile-away-class-properties-since-browsers-support
And a minimal babel repro here: https://github.com/NullVoxPopuli/babel-transpilation-tests
I expect that setting my targets to `last 1 firefox versions` would allow me to keep native class properties. and only transpile what is decororated.

I believe fixing this could provide a bit of a boost to everyone's build times. Today all class properties are transpiled to a time when class properties were not implemented in the browser).

### 🔬 Minimal Reproduction

 - ember new my-app
 - cd my-app
 - remove safari from targets, leaving only `last 1 Chrome versions` and `last 1 firefox versions`
 - add browserslist to package.json
 - add browserslist as devDependency 
 - `yarn browserslist --update-db`
 - run `ember build -e production`
 - in the my-app.js, I expect all occurrences of `rootURL` to be involved in an assignment, rather than a function call.
 
 result: https://github.com/NullVoxPopuli/repro-for-emberjs-19790
 
### 😕 Actual Behavior
Living in the dark ages

### 🤔 Expected Behavior
Utilize support for native properties, private properties, methods, etc.

### 🌍 Environment

- Ember: - CLI 3.28.0
- Node.js/npm: -14.17.6
- OS: - n/a
- Browser: - n/a

### ➕ Additional Context
See the comparison between these two files:

Input: https://github.com/NullVoxPopuli/babel-transpilation-tests/blob/main/input.js
Output: https://github.com/NullVoxPopuli/babel-transpilation-tests/blob/main/output.js

This is correct.
But when I add that input file to ember, I get this for the output: 
```
/home/nullvoxpopuli/Development/tmp/my-app/my-app/input.js: Class private methods are not enabled.
  2 |   #a = 'hi';
  3 |
> 4 |   get #b() {
    |   ^
  5 |     return 'hello' + this.#a;
  6 |   }
  7 |
```
which... I guess is unrelated, but should still be fixed. Now, If I change the private methods/getters to public:
I get this mess:
```js
  var e, t, r
[...]

PR #449, opened 2022-06-26
Avoid transpiling class-properties as they are natively supported

NullVoxPopuli:
**This PR is incomplete** -- current status: babel problem

_goal_:  make babel do less:
tested at: https://github.com/NullVoxPopuli/babel-transpilation-tests
current problem: (all in babel, not ember-cli-babel) decorator plugin assumes class-properties plugin is needed (it's not)

This is not intended to be merged, but serves as something for folks to point their package.jsons at when they want to see if they can disable class-property transpilation across their whole app and addons.

hopefully: _class property transformation is not required for decorators_ .

(it used to be required tho).
There is an assertion / error that's added by the decorators plugin tho.

Upon manual removal in the above demo _everything still works fine_.

Here is the the test:

<details><summary>input</summary>

```js
function f(target, key, desc) {
  let oldGet = desc.get
  let oldInit = desc.initializer

  desc.get =() => {
    console.log('called f on ', key);
    return oldInit?.() || oldGet();
  }
}
class A {
  #a = 'hi';

  get #b() {
    return 'hello' + this.#a;
  }

  get c() { return this.#b; }

  @f g;

  @f h = 2;

  i = 3;
}

let a = new A();

console.log({ g: f.g, c: a.c, h: a.h, i: a.i });
```

</details>
<details><summary>output</summary>

remove lines which define g and h
```js
"use strict";

var _class, _descriptor, _descriptor2;

function _applyDecoratedDescriptor(target, property, decorators, descriptor, context) { var desc = {}; Object.keys(descriptor).forEach(function (key) { desc[key] = descriptor[key]; }); desc.enumerable = !!desc.enumerable; desc.configurable = !!desc.configurable; if ('value' in desc || desc.initializer) { desc.writable = true; } desc = decorators.slice().reverse().reduce(function (desc, decorator) { return decorator(target, property, desc) || desc; }, desc); if (context && desc.initializer !== void 0) { desc.value = desc.initializer ? desc.initializer.call(context) : void 0; desc.initializer = undefined; } if (desc.initializer === void 0) { Object.defineProperty(target, property, desc); desc = null; } return desc; }

function _initializerWarningHelper(descriptor, context) { throw new Error('Decorating class property failed. Please ensure that ' + 'proposal-class-properties is enabled and runs after the decorators transform.'); }

function f(target, key, desc) {
  let oldGet = desc.get;
  let oldInit = desc.initializer;

  desc.get = () => {
    console.log('called f on ', key);
    return oldInit?.() || oldGet();
  };
}

let A = (_class = class A {
  #a = 'hi';

  get #b() {
    return 'hello' + this.#a;
  }

  get c() {
    return this.#b;
  }

  g = _initializerWarningHelper(_descriptor, this);
  h = _initializerWarningHelper(_descriptor2, this);
  i = 3;
}, (_descriptor = _applyDecoratedDescriptor(_class.prototype, "g", [f], {
  configurable: true,
  enumerable: true,
  writable: true,
  initializer: null
}), _descriptor2 = _applyDecoratedDescriptor(_class.prototype, "h", [f], {
  configurable: true,
  enumerable: true,
  writable: true,
  initializer: function () {
    return 2;
  }
})), _class);
let a = new A();
console.log({
  g: f.g,
  c: a.c,
  h: a.h,
  i: a.i
});

```

</details>

the goal of this effort is the following:
 - to reduce the amount of work babel is doing 
 - to use _one_ babel / ember-cli-babel instance for the whole app tree (the complexity brought by addons compliing themselves with their own ember-cli-babel has been problematic for app-land control).
 
 To do this, a consuming app must have _only one_ ember-cli-babel instance in the whole node_modules graph.
 
 Related issues:
  - https://github.com/babel/ember-cli-babel/issues/447
  - https://github.com/babel/ember-cli-babel/issues/446
  - https://github.com/babel/ember-cli-babel/issues/417
  - https://github.com/babel/ember-cli-babel/issues/447

### DIFF
--- a/lib/babel-options-util.js
+++ b/lib/babel-options-util.js
@@ -321,28 +321,8 @@ function _addDecoratorPlugins(plugins, options, config, parent, project) {
   } else {
     addPlugin(
       plugins,
-      [require.resolve("@babel/plugin-proposal-decorators"), { legacy: true }],
-      _buildClassFeaturePluginConstraints(
-        {
-          before: ["@babel/plugin-proposal-class-properties"],
-        },
-        config,
-        parent,
-        project
-      )
+      [require.resolve("@babel/plugin-proposal-decorators"), { legacy: true }]
     );
-  }
-
-  if (hasPlugin(plugins, "@babel/plugin-proposal-class-properties")) {
-    if (parent === project) {
-      project.ui.writeWarnLine(
-        `${_parentName(
-          parent
-        )} has added the class-properties plugin to its build, but ember-cli-babel provides these by default now! You can remove the transforms, or the addon that provided them, such as @ember-decorators/babel-transforms.`
-      );
-    }
-  } else {
-    _addClassProperties(addPlugin, plugins, options, config, parent, project);
   }
 
   if (hasPlugin(plugins, "babel-plugin-filter-imports")) {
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
This won't be possible. best place we can spend effort is moving to platform-decorators

Issue #442, opened 2022-04-25
Private property transpilation broken w/ externalHelpers enabled


nickschot:
ember-cli-babel currently has a pinned dependency of `@babel/runtime` 7.12.18. If a newer version (like 7.17.9) for whatever reason ends up on the root level of node_modules (like other dependencies bringing in a newer version) it may break the build at runtime in certain cases. The specific case that I ran into occurs when using private properties in native classes like in this reproduction: https://github.com/nickschot/babel-test-app/blob/master/app/controllers/application.js#L9 

with `externalHelpers: true` configured [in ember-cli-build](https://github.com/nickschot/babel-test-app/blob/master/ember-cli-build.js#L9).

When running the app this will result in a runtime error like:
<img width="1278" alt="Screenshot 2022-04-25 at 16 06 39" src="https://user-images.githubusercontent.com/334789/165105812-3b758d63-2cc2-4e22-bb72-8c847d12ba74.png">

The reproduction uses a private property in an app controller, but the same issue happens when addons use private properties in for example components.

Note that for the linked reproduction I've added a dependency on `@babel/runtime` 7.17.9 to force this issue. Disabling `externalHelpers` and/or making sure the version of `@babel/runtime` in the root of `node_modules` is 7.12.18 prevents the issue from occurring.

The pinning happened here: https://github.com/babel/ember-cli-babel/pull/385
May be somewhat related to: https://github.com/babel/ember-cli-babel/pull/420

PR #451, opened 2022-07-06
unpin @babel/runtime to see what tests fail

mansona:
This is just unpinning `@babel/runtime` to see what tests fail so that we might be able to get to the bottom of the issues with https://github.com/babel/ember-cli-babel/issues/442


boris-petrov:
@mansona did you get anywhere with this? I just noticed that in 8.2.0 the version is still pinned.

PR #452, opened 2022-08-10
Move `@babel/core` to `peerDependencies` to resolve peer dependency warnings and errors

NullVoxPopuli:
…confusion

Following the advice here: https://github.com/ember-cli/ember-cli/pull/9934#issuecomment-1210078896

The eventual perceived fix for users is that when using strict package managers (npm 8, yarn2+, pnpm, etc) is that when they are told to specify a missing peer, `@babel/core`, the `@babel/core` they specify will be the one used by ember-cli-babel.

`@babel/core` will eventually be added to the app and addon blueprints in https://github.com/ember-cli/ember-cli/pull/9934

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,8 +40,10 @@
     "test:node": "mocha node-tests",
     "test:node:debug": "mocha debug node-tests"
   },
+  "peerDependencies": {
+    "@babel/core": "^7.12.0"
+  },
   "dependencies": {
-    "@babel/core": "^7.12.0",
     "@babel/helper-compilation-targets": "^7.12.0",
     "@babel/plugin-proposal-class-properties": "^7.16.5",
     "@babel/plugin-proposal-decorators": "^7.13.5",
@@ -73,6 +75,7 @@
     "semver": "^5.5.0"
   },
   "devDependencies": {
+    "@babel/core": "^7.12.0",
     "@glimmer/component": "^1.0.4",
     "@glimmer/tracking": "^1.0.4",
     "babel-eslint": "^10.1.0",


NullVoxPopuli:
I have 8 failures when I run the tests -- but those 8 failures also exist on `master` so.... that's fun

ef4:
You need a devDependency too. 

(This is an example of the confusing fallout from making v1 addons be their own test-app too, in a single package. The package.json has two meanings. It needs to do the work that a consuming app would do (put `@babel/core` in devDependencies) while also doing the work of the addon (put `@babel/core` in peerDependencies).

NullVoxPopuli:
devDependency is there

ef4:
Oops, sorry, somehow I looked right at it and didn't see it. 😛

bertdeblock (approved):

ef4:
I think we can make the argument that this is a bugfix instead of breaking change because people were *already* experiencing peer dep warnings and/or failures depending on their package manager, so documenting the fact that you already needed this peer dep to avoid those issues is only a bug fix.

bertdeblock:
This seems ready to go?

ef4:
It was pointed out to me that this really is probably breaking, because it's not the addition to `peerDependencies` that is the big problem, it's the removal from `dependencies`.

ef4:
I still think we should do it, we can make this ember-cli-babel 8.

I know a traditional reason not to do a major release here was to save it for a babel major release, but this change means we're decoupled from the babel version anyway.

NullVoxPopuli:
Sounds good to me -- as long as we can specify in the release notes that `resolutions` / `overrides` can still be used to collapse all ember-cli-babel versions down to 1 version in folks dep-trees, and v8 would still be compatible with their projects, provided they add `@babel/core` (and likely meet other criteria we move as a part of the major (node engines, etc))

ef4:
Discussed with ember-cli core team and agreed to merge this, once we see it run CI. For some reason that hasn't happened yet. Investigating.

Also, the plan to release this is tracked in https://github.com/babel/ember-cli-babel/issues/453 which I will fill out with more details.

bertdeblock:
@NullVoxPopuli I added a test commit and drop it again to trigger CI, hope you don't mind!

bertdeblock:
CI is green, so I'm merging!

Issue #453, opened 2022-09-16
[QUEST] v8 Release


bertdeblock:
- [x] Move `@babel/core` to `peerDependencies` - https://github.com/babel/ember-cli-babel/pull/452
- [x] Drop support for Node v6, v8, v10 and v12 (Add support for Node v14, v16 and v18) - https://github.com/babel/ember-cli-babel/pull/454
- [x] Remove `includePolyfill` support - https://github.com/babel/ember-cli-babel/pull/458
- [x] Change `broccoli-babel-transpiler` to either make `@babel/core` a peer dependency or make it accept the current Babel as an argument, and upgrade `ember-cli-babel` to that version (without this we're still not actually transpiling people's code with the version of `@babel/core` in their own `package.json`) - https://github.com/babel/broccoli-babel-transpiler/pull/221
- [x] Update `broccoli-babel-transpiler` to v8 - https://github.com/babel/ember-cli-babel/pull/460
- [x] Write upgrade guide for apps and addons - https://github.com/babel/ember-cli-babel/pull/464
- [x] Release first v8 beta version
- [x] Release a stable version of `broccoli-babel-transpiler` v8 and update in `ember-cli-babel` - https://github.com/babel/ember-cli-babel/pull/494
- [x] Release a stable version of `ember-cli-babel` v8
- [x] Update `ember-cli-babel` in the default blueprint and add `@babel/core` - https://github.com/ember-cli/ember-cli/pull/10141

kategengler:
from @rwjblue in Discord
> rwjblue — 09/16/2022
Agree
There’s a non trivial amount of cleanup we should do though
Like bumping the broccoli-Babel-transpiler version
Changing poly fill support around (removing it basically)

ef4:
Also just noticed that broccoli-babel-transpiler has `@babel/core` in `dependencies`. That means we should also change and release broccoli-babel-transpiler to either use a peerDep *or* accept babel itself as an argument when you create the broccoli node.

ef4:
(I can't edit the description above, so I'm rewriting it here with updates.)

 - [ ] move `@babel/core` to peerDependencies https://github.com/babel/ember-cli-babel/pull/452
 - [ ] drop support for node < 14 and add test coverage for 14 through 18
 - [ ] remove `includePolyfill` support
[...]

PR #458, opened 2022-10-01
Drop support for the `includePolyfill` option

bertdeblock:
[@babel/polyfill](https://babeljs.io/docs/en/babel-polyfill) is deprecated, the replacement for users would be to include `core-js/stable` directly:

```js
// app.js

import 'core-js/stable';
```

Builds on #454, as running `ember test` wasn't working for me locally, because of the macOS + Ember CLI issue.

Related to #453.


rwjblue (approved):

package.json
```diff
@@ -52,7 +52,6 @@
     "@babel/plugin-transform-modules-amd": "^7.13.0",
     "@babel/plugin-transform-runtime": "^7.13.9",
     "@babel/plugin-transform-typescript": "^7.13.0",
-    "@babel/polyfill": "^7.11.5",
```
SergeAstapov, line 55: this is awesome!

PR #461, opened 2022-10-31
Update to the v4.8.0 addon blueprint

Windvis:


README.md
```diff
@@ -37,7 +37,9 @@ ember install ember-cli-babel
 
 ## Compatibility
 
-- ember-cli-babel 7.x requires ember-cli 2.13 or above
+* Ember.js v3.28 or above
```
Windvis, line 40: To be discussed what this needs to be, but it seems we don't run any ember-try scenarios yet?

PR #467, opened 2022-11-24
Clarify how to set `enableTypeScriptTransform` in Ember apps, addons and engines

ijlee2:
## Description

Currently, the `README` explains how to set `enableTypeScriptTransform` for Ember apps only.

It wasn't clear to me what to do for Ember addons and engines (to be precise, [engines created as an addon](https://ember-engines.com/docs/quickstart#create-as-addon)). The setup for engines turned out to be tricky, because I wasn't supposed to embed the `'ember-cli-babel'` key inside of an `options` hash (the pattern suggested by apps and addons).

I'd like to update the section `Enabling TypeScript Transpilation` so that:

- A developer can copy-paste the code example.
- The code examples closely match the latest Ember blueprint (`v4.9.0-beta.0` at the time of writing). (Note, I continued to write `return app.toTree();`. That is, to limit the scope of this pull request, I didn't assume that the developer uses, or will use, Embroider.)
- The surrounding texts have shorter sentences and are easier to understand. (The change may help those whose native language is not English.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,11 +351,6 @@ NOTE: Setting `enableTypeScriptTransform` to `true` does *not* enable
 type-checking. For integrated type-checking, you will need
 [`ember-cli-typescript`](https://ember-cli-typescript.com).
 
-NOTE: Setting `enableTypeScriptTransform` to `true` is not compatible with
-`ember-cli-typescript < 4.0` because of conflicting Babel plugin ordering
-constraints. It is also unnecessary because `ember-cli-typescript < 4.0`
-adds the TypeScript-Babel transform.
-
 ### Babel config usage
 
 If you want to use the existing babel config from your project instead of the auto-generated one from this addon, then you would need to *opt-in* by passing the config `useBabelConfig: true` as a child property of `ember-cli-babel` in your `ember-cli-build.js` file.
--- a/README.md
+++ b/README.md
@@ -242,35 +242,120 @@ module.exports = function(defaults) {
 
 #### Enabling TypeScript Transpilation
 
-The transform plugin required for Babel to transpile TypeScript will
-automatically be enabled when `ember-cli-typescript` >= 4.0 is installed.
+Babel needs a transform plugin in order to transpile TypeScript. When you
+install `ember-cli-typescript >= 4.0`, this plugin is automatically enabled.
 
-You can enable the TypeScript Babel transform manually *without*
-`ember-cli-typescript` by setting the `enableTypeScriptTransform` to `true`.
+If you don't want to install `ember-cli-typescript`, you can still enable
+the TypeScript-Babel transform. You will need to set `enableTypeScriptTransform`
+to `true` in select file(s).
 
-NOTE: Setting this option to `true` is not compatible with
-`ember-cli-typescript` < 4.0 because of conflicting Babel plugin ordering
-constraints and is unnecessary because `ember-cli-typescript` < 4.0 adds the
-TypeScript Babel transform itself.
 
-NOTE: Setting this option to `true` does *not* enable type-checking. For
-integrated type-checking, you will need
-[`ember-cli-typescript`](https://ember-cli-typescript.com).
-
-In an app, manually enabling the TypeScript transform would look like:
+<details>
+<summary>Apps</summary>
 
 ```js
-// ember-cli-build.js
-module.exports = function(defaults) {
-  let app = new EmberApp(defaults, {
+/* ember-cli-build.js */
+
+const EmberApp = require('ember-cli/lib/broccoli/ember-app');
+
+module.exports = function (defaults) {
+  const app = new EmberApp(defaults, {
+    // Add options here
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -242,35 +242,120 @@ module.exports = function(defaults) {
...
+    enableTypeScriptTransform: true,
+  },
+});
+```
+
+</details>
+
+
+NOTE: Setting `enableTypeScriptTransform` to `true` does *not* enable
+type-checking. For integrated type-checking, you will need
```
NullVoxPopuli, line 351: Idk if we want to mention this, given the typed ember team's plans to eventually sunset ec-typescript.

@chriskrycho thoughts?

ijlee2: Ah. Thanks for providing additional context.

Both `NOTE`'s are shown currently in the `README` (https://github.com/babel/ember-cli-babel/tree/v7.26.11#enabling-typescript-transpilation).  For this reason, I'd suggest that we remove references to `ember-cli-typescript` in a separate task, when its support is ended. In this pull request, I only wanted to show code examples and simplify the surrounding texts.

chriskrycho: Yeah, we shouldn't update the info about ember-cli-typescript until it actually changes!

README.md
```diff
@@ -242,35 +242,120 @@ module.exports = function(defaults) {
...
+
+module.exports = function (defaults) {
+  const app = new EmberAddon(defaults, {
+    // Add options here
+    'ember-cli-babel': {
+      enableTypeScriptTransform: true,
+    },
+  });
+
+  // ...
```
ijlee2, line 329:
```suggestion
  return app.toTree();
```

README.md
```diff
@@ -242,35 +242,120 @@ module.exports = function(defaults) {
...
+
+NOTE: Setting `enableTypeScriptTransform` to `true` does *not* enable
+type-checking. For integrated type-checking, you will need
+[`ember-cli-typescript`](https://ember-cli-typescript.com).
+
+NOTE: Setting `enableTypeScriptTransform` to `true` is not compatible with
+`ember-cli-typescript < 4.0` because of conflicting Babel plugin ordering
+constraints. It is also unnecessary because `ember-cli-typescript < 4.0`
+adds the TypeScript-Babel transform.
+
```
chriskrycho, line 358: I think we can just drop this paragraph at this point; 4.0 came out fall 2020, and the last 3.x release was May 2020, so I don't think it's particularly relevant anymore.

```suggestion
```

chriskrycho (approved):

ijlee2:
@chriskrycho When `ember-cli-typescript` is installed in a `v1` addon, two scripts get added to `package.json`:

```json
{
  "scripts": {
    "prepack": "ember ts:precompile",
    "postpack": "ember ts:clean"
  }
}
```

When we use `enableTypeScriptTransform` and uninstall `ember-cli-typescript`, is it okay to remove these two scripts (i.e. the addon and its types will be published without an issue)?

If so, I wonder if this is worth mentioning in `ember-cli-babel`'s `README`. I can also see the perspective that it may fit better in a migration guide in the `ember-cli-typescript` repo.

chriskrycho:
Nope, nothing will generate types out of the box for you in that mode. For that reason, addons which want to publish types should generally do one of two things:

- Classic addons should continue to use `ember-cli-typescript`.
- Modern/Embroider addons should use the TS-ready rollup config for addons that @simonihmig and others built, which will publish types correctly.

Probably worth documenting that explicitly here (feel free to steal that prose and tweak/improve it!).

README.md
```diff
@@ -242,35 +242,115 @@ module.exports = function(defaults) {
...
+  return app.toTree();
+};
+```
+
+```js
+/* index.js */
+
+const EngineAddon = require('ember-engines/lib/engine-addon');
+
+module.exports = EngineAddon.extend({
```
bertdeblock, line 338: In [v0.8.13](https://github.com/ember-engines/ember-engines/blob/master/CHANGELOG.md#v0813-2021-04-14), a `buildEngine` function was added to replace `extend`. It's also used in the docs, for example [here](https://ember-engines.com/docs/lazy-loading).

bertdeblock (commented):
AFAICT, this looks good @ijlee2.

One thing I'm wondering though, I'm not sure if the `ember-cli-build.js` file for addons and engines should be modified as well. I think that enabling the transform from within the `index.js` file should be sufficient.

README.md
ijlee2: @bertdeblock Thanks for the suggestion. I wasn't familiar with the new syntax and hope I updated the file correctly.

backspace:
Thanks for writing this up, it helped me when I was having build problems with an addon. I think it would be great to have merged so it’s easier to find 😀

ijlee2:
@bertdeblock Could we merge this pull request some time soon? (I forget where we had left off, if there had been additional feedback that I needed to address.)

bertdeblock (approved):

PR #468, opened 2022-11-29
Allow passing in generatorOpts

colenso:
Allows passing in `generatorOpts` to babel  https://babeljs.io/docs/en/options#generatoropts without passing it to preset-env. This will allow passing options like `compact` etc to babel.
Did this as referenced in https://github.com/ef4/ember-auto-import/issues/549#issuecomment-1330104695

### DIFF
--- a/index.js
+++ b/index.js
@@ -113,6 +113,7 @@ module.exports = {
     let providedAnnotation;
     let throwUnlessParallelizable;
     let sourceMaps = false;
+    let generatorOpts = {};
     let shouldCompileModules = _shouldCompileModules(config, this.project);
 
     if (emberCLIBabelConfig) {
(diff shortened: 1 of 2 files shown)


ef4:
Thanks, this looks good to me. Not having it is definitely a gap, and telling people to switch to `useBabelConfig: true` (as mentioned in other issues about this) without creating an upgrade path for their addon-provided babel plugins isn't really a viable alternative at this time.

ef4:
The 8.0.0 release is [imminent](https://github.com/babel/ember-cli-babel/issues/453), so this will be in there. I don't think 8.0 will be a difficult upgrade for most people, but if it turns out to be we can also consider back-porting this.

colenso:
> 

Awesome! Thanks

PR #489, opened 2023-07-06
remove proposal plugins and replace with their corresponding transform

mansona:
plugin-proposal-class-properties, plugin-proposal-decorators, plugin-proposal-private-methods, and plugin-proposal-private-property-in-object are all now renamed to transform since they are out of proposal stage so this PR updates the dependencies 👍


mansona:
So I noticed when the tests were failing that actually I missed some code that was trying to do smart stuff with ordering the plugins. I have pushed a really naive way to update that code but I have a seeking suspicion that we need to check for both the cases of "proposal" plugins and "transform" plugins, favouring the transform ones but catering for the case that people still use proposal ones 🤔 

Can someone with more context in this area let me know if they thing that is required?

Windvis:
I _think_ we can simply remove these plugins as they are part of preset-env: #501

PR #491, opened 2023-08-08
Add note on TS in adding to README.md

chriskrycho:
We have documented the use of `enableTypeScriptTransform` for a long time, but only for apps. Add the corresponding note for addons, in a way which should also help for searchability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,25 @@ module.exports = function(defaults) {
   return app.toTree();
 }
 ```
+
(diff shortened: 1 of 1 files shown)


locks (approved):

NullVoxPopuli (approved):

bertdeblock:
I merged https://github.com/babel/ember-cli-babel/pull/491, so I think we can close this one if that's okay.

PR #493, opened 2023-08-17
Drop support for Node v14 and add support for Node v20

bertdeblock:
Might as well include this in the upcoming v8 release.


kategengler:
Thank you!

PR #500, opened 2023-08-27
fix: filter undefined options

### LINKED ISSUE #499: ember-cli-babel@8 breaks parallelization
This PR https://github.com/babel/ember-cli-babel/pull/462
Which partially reverted this PR https://github.com/babel/ember-cli-babel/pull/460

Falls victim to a common issue with destructuring that is unsafe for broccoli-babel-transpiler because broccoli-babel-transpiler treats `undefined` as unserializable.

Specifically these lines need to account for undefined values https://github.com/bertdeblock/ember-cli-babel/blob/188d15abcba3830dd7ca0b22abe10453d5a6fea3/index.js#L151-L175

runspired:
resolves #499 

PR with a patch demonstrating this fix for EmberData https://github.com/emberjs/data/pull/8800

### DIFF
--- a/index.js
+++ b/index.js
@@ -181,6 +181,14 @@ module.exports = {
       },
     };
 
+    // Remove any undefined options so that they don't override the default
+    // Or error when broccoli-babel-transpiler checks serializability
+    Object.keys(options.babel).forEach(key => {
+      if (options.babel[key] === undefined) {
+        delete options.babel[key];
+      }
+    });
+
     let output;
 
     const customAddonConfig = config['ember-cli-babel'];


runspired:
looks like there was a dependency installation timeout on the two failing scenarios and they just need a kick.

NullVoxPopuli:
Any chance you'd be willing to add a test for this? in particular, around here: https://github.com/babel/ember-cli-babel/blob/master/node-tests/addon-test.js#L1689

bertdeblock (approved):

PR #502, opened 2023-09-23
Ensure that plugin-transform-class-static-block is provided

NullVoxPopuli:
Found this issue while working on: https://github.com/ember-template-imports/ember-template-imports/pull/187

I tested over on that branch locally by editing the node_modules directly.

The test suite here in ember-cli-babel can't be trusted because `@babel/plugin-transform/class-static-block` was picked up automatically due to how bad yarn@v1's hoisting is -- the static-block plugin was resolved via `@babel/preset-env` -- preset-env does not inherently provide static-blocks -- it _must_ be configured when we compile away class properties.

If this repo were to switch to using a real package manager, the the static test would fail without the corresponding change in `lib/ember-plugins.js`


NullVoxPopuli:
failures look to be due to network issues

ef4 (approved):
This looks good to me.

It's yet another required transform when using legacy decorators, so not really any new policy choice here.

While we talked about getting this into 8.0.1 (which just shipped),  I think it's better that we wanted so this can be a minor instead, as this adds a new capability.

PR #503, opened 2023-10-03
Support static class blocks when _not_ using an explicit babel config file.

NullVoxPopuli:
In https://github.com/babel/ember-cli-babel/pull/502
Support for static blocks was added only when using an explicit babel config.

This PR adds the implicit / default support that most folks would use in today's ember apps.

I had missed this because I didn't, at the time, know that the paths for generating which babel plugins we need between the two setups (with and without a babel config) use separate code.

Unblocks: https://github.com/ember-template-imports/ember-template-imports/pull/187

### DIFF
--- a/lib/babel-options-util.js
+++ b/lib/babel-options-util.js
@@ -316,7 +316,7 @@ function _addDecoratorPlugins(plugins, options, config, parent, project) {
       project.ui.writeWarnLine(
         `${_parentName(
           parent
-        )} has added the static class block plugin to its build, but ember-cli-babel provides these by default now! You can remove the transforms, or the addon that provided them, such as @ember-decorators/babel-transforms. Ember supports the stage 1 decorator spec and transforms, so if you were using stage 2, you'll need to ensure that your decorators are compatible, or convert them to stage 1.`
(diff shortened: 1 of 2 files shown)


lib/babel-options-util.js
```diff
@@ -311,6 +311,29 @@ function _buildClassFeaturePluginConstraints(constraints, config, parent, projec
 function _addDecoratorPlugins(plugins, options, config, parent, project) {
   const { hasPlugin, addPlugin } = require("ember-cli-babel-plugin-helpers");
 
+  if (hasPlugin(plugins, "@babel/plugin-transform-class-static-block")) {
+    if (parent === project) {
+      project.ui.writeWarnLine(
+        `${_parentName(
+          parent
+        )} has added the static class block plugin to its build, but ember-cli-babel provides these by default now! You can remove the transforms, or the addon that provided them, such as @ember-decorators/babel-transforms. Ember supports the stage 1 decorator spec and transforms, so if you were using stage 2, you'll need to ensure that your decorators are compatible, or convert them to stage 1.`
```
NullVoxPopuli, line 319: todo: update this message

ef4 (approved):

PR #506, opened 2023-10-24
Add import.meta.env support

NullVoxPopuli:
This'll be required to consume glimmer-vm as it is published today.

Work spawned from: https://github.com/emberjs/ember.js/pull/20561


NullVoxPopuli:
gonna pursue an alternate route where glimmer-vm packages use proper conditional exports (package.json config)

PR #530, opened 2026-02-04
Upgrade babel deps

NullVoxPopuli:
- @babel/*
  - just some patches (and vuln fixes)

- babel-plugin-debug-macros --> skipped to v2 -- in range for current support
  - drop babel < 7
  - drop node < 16


package.json
```diff
@@ -40,19 +40,19 @@
...
+    "@babel/plugin-proposal-decorators": "^7.29.0",
+    "@babel/plugin-transform-class-properties": "^7.28.6",
+    "@babel/plugin-transform-class-static-block": "^7.28.6",
+    "@babel/plugin-transform-modules-amd": "^7.27.1",
+    "@babel/plugin-transform-private-methods": "^7.28.6",
+    "@babel/plugin-transform-private-property-in-object": "^7.28.6",
+    "@babel/plugin-transform-runtime": "^7.29.0",
+    "@babel/plugin-transform-typescript": "^7.28.6",
+    "@babel/preset-env": "^7.29.0",
+    "@babel/runtime": "7.28.6",
```
mansona: shouldn't this have a `^` on it?

NullVoxPopuli: freeeeeeakin pnpm up

lib/relative-module-paths.js
```diff
@@ -11,6 +11,7 @@ function getRelativeModulePath(modulePath) {
 }
 
 function resolveRelativeModulePath(name, child) {
+  name = name.replace(/\.js$/, '');
```
NullVoxPopuli: this is needed because `@babel/runtime` is a `"type": "commonjs"` package that uses extensions -- which node supports but broccoli does not.

package.json
NullVoxPopuli: oh actually pnpm did the right thing, it's just not what we want -- it was previously pinned 🙈

lib/relative-module-paths.js
void-mAlex: is that still true for babel8?

NullVoxPopuli: babel 8 is ESM only, so I don't know if it would be compatible with broccoli -- maybe more hacks needed

mansona: the latest broccoli only supports node 20.19 because it relies on `require(esm)`. We're at the point where we can start doing the same in the wider ecosystem. This might be worth discussing at the next tooling team meeting

Also there is nothing stopping cjs from supporting extensions 🤔 what is the specific error here that the change is trying to fix?

NullVoxPopuli: here is what the failure looks like: https://github.com/emberjs/ember-cli-babel/pull/531

NullVoxPopuli:
```
not ok 1 Chrome 144.0 - [undefined ms] - Global error: Uncaught Error: Could not find module `@babel/runtime/helpers/esm/toPropertyKey.js` imported from `@babel/runtime/helpers/esm/defineProperty` at http://localhost:7357/assets/vendor.js, line 266
    ---
        browser log: |
            {"type":"error","text":"Uncaught Error: Could not find module `@babel/runtime/helpers/esm/toPropertyKey.js` imported from `@babel/runtime/helpers/esm/defineProperty` at http://localhost:7357/assets/vendor.js, line 266\n","testContext":{}}
```

the import/require line from a define:
image

all defines have no extensions tho:
image

NullVoxPopuli: https://discord.com/channels/480462759797063690/1468675776760119400/1468691870824534097

NullVoxPopuli: > Also there is nothing stopping cjs from supporting extensions

CJS supports it yes.
It's really AMD / loader.js that does not.

mansona: this looks very much like the wrong level to do this and will probably be very brittle

It looks like the babel-plugin that we're using supports passing either `extensions` or `stripExtensions` and that seems like it would be the more stable way forward: https://github.com/tleunen/babel-plugin-module-resolver/blob/master/DOCS.md#extensions

NullVoxPopuli: image

js is already a default extension


https://github.com/tleunen/babel-plugin-module-resolver/blob/master/src/normalizeOptions.js#L11

I'll debug it, but this makes me despair lol

NullVoxPopuli: no, thee fix does not go in babel-plugin-module-resolver.

we override resolving logic with the function I changed -- so this is our own doing, and our own fix.

There is this: https://npmx.dev/package/amd-name-resolver which I also think is the wrong place to the fix.

The only place I think a deeper fix makes sense is loader.js.

_or_ we lean in to composition, and because we have this need to eliminate extensions (and we don't want to encourage the broader ecosystem to use AMD), we just fix it here in ember-cli-babel.

The whole of `moduleResolve` could probably go away tho... because we have require resolve, and the URL utils.

but that feels like a bigger effort for another PR, rather than a dep-upgrade PR

NullVoxPopuli: tl;dr: I think fixing in ember-cli-babel is correct

NullVoxPopuli: _especially_ since the alternative is to reach in to private code and call babel-plugin-module-resolver's own stripExtensions:
```js
function stripExtension(modulePath, stripExtensions) {
  let name = _path.default.basename(modulePath);
  stripExtensions.some(extension => {
    if (name.endsWith(extension)) {
      name = name.slice(0, name.length - extension.length);
      return true;
    }
    return false;
  });
  return name;
}
```

which is about as safe as what I have in this PR -- albeit likely slower

NullVoxPopuli: I did PR this to the upgrade branch tho: https://github.com/emberjs/ember-cli-babel/pull/532

ef4:
From tooling team discussion:

A more targeted fix is possible here: https://github.com/emberjs/ember-cli-babel/blob/bdb4bd14853630dcc052ed2a7ed5745309eaa264/index.js#L271

because the contents of `@babel/runtime` that end up in the AMD modules get there entirely because of that code. We can patch/adjust babel/runtime so that the imports and the defines line up correctly.

NullVoxPopuli:
what do you mean? what is there to configure/rename with that code?

> We can patch/adjust babel/runtime so that the imports and the defines line up correctly.

I don't understand what you mean

mansona:
> I don't understand what you mean

This is the more targeted fix that Ed was talking about: https://github.com/emberjs/ember-cli-babel/pull/538

NullVoxPopuli:
have fun learning all the context :p 

(wouldn't pinning also prevent folks from using babel 8 fully?)

mansona:
> wouldn't pinning also prevent folks from using babel 8 fully?

I really don't understand the question 🙈 The PR that I linked **unpins** `@babel/runtime` which should be the first step necessary to start supporting both babel 7 and babel 8.

NullVoxPopuli:
this hasn't come up in a while and I don't have any motivation to follow up with this, so as a part of cleaning up my PRs so my [PR view](https://github.com/search?q=is%3Apr+author%3ANullVoxPopuli++is%3Aopen+-is%3Adraft++org%3Aember-learn++org%3Aembroider-build++org%3Aemberjs++org%3Aember-cli++org%3Aember-tooling+org%3Arelease-plan+org%3Amainmatter&type=pullrequests&s=updated&o=desc&p=10) is more focused on what I have energy for, I am not misled by things I'm not actually going to follow up with.

PR #532, opened 2026-02-05
doing what babel-plugin-module-resolver would do

NullVoxPopuli:
Extra code for removing `.js` from file paths so that AMD / loader.js works

In support or defense of https://github.com/emberjs/ember-cli-babel/pull/530


NullVoxPopuli:
i don't really have time to follow up wit this, and this problem / need for this has not come up since moving away from ember-cli-babel, so I'm going to close so that I can focus on other things

Issue #519, opened 2025-03-12
Does `@babel/runtime` still need pin?


gorner:
`@ember/runtime` was pinned to v7.12.8 in https://github.com/emberjs/ember-cli-babel/pull/385 (March 2021) due to "workaround test failures in CI for the "Floating dependencies" scenario."

Is this pin still needed / relevant? As it stands, this appears to block updating `@ember/runtime` to v7.26.10 in our app (see https://github.com/babel/babel/security/advisories/GHSA-968p-4wvh-cqc8).

gorner:
I briefly looked at opening a PR myself to at least change the `@babel/runtime` dependency to `"^7.26.10"` but ran into issues:

- If I only change the one dependency, `mocha node-tests` passes but `ember test` fails with:
    ```
    Uncaught Error: Could not find module `@babel/runtime/helpers/esm/toPropertyKey.js` imported from `@babel/runtime/helpers/esm/defineProperty`
    ```
- If I remove yarn.lock and reinstall dependencies from scratch, one mocha test (`can compile static blocks`) fails, and `ember test` still fails for the same reason as before.

Using npm overrides in an Ember app with `"@babel/runtime": "^7.26.10",` does not seem to cause any issues in the app I work on, at least after some quick smoke testing.

However I see now that there is an existing issue #442 suggesting that there are/were some other compatibility issues preventing it from being unpinned before now.

gorner:
So I've figured out a potential fix, at least in POC form. But as far as I can tell, it requires a change to https://github.com/ember-cli/loader.js to allow imports with a `.js` extension to resolve to the correct module that was loaded without that extension. 

I have a draft PR with that change here: https://github.com/ember-cli/loader.js/pull/227 — but it only handles the `.js` case and might need to be more robust. Any guidance from the Ember/CLI core teams on how to proceed or if it's worth pursuing this would be welcome.

gorner:
OK, thanks @ef4. So if I'm understanding correctly, the solution to clearing the advisory is:

1. ensure your app/addons don't use `includeExternalHelpers` (I've confirmed mine do not)
2. follow the steps in [your comment on #442](https://github.com/emberjs/ember-cli-babel/issues/442#issuecomment-2894738580) to depend directly on `@babel/runtime`
3. use NPM overrides or equivalent if necessary to clear the old version from the build tree. 

Correct?

Issue #445, opened 2022-05-23
`Could not find module '@babel/runtime/helpers/esm/regeneratorRuntime'` in a fresh new app


boris-petrov:
Run `ember new --yarn ember-cli-babel-bug`. Then modify `config/targets.js` to be just `['firefox 35']`. To `ember-cli-build.js` add:

```js
'ember-cli-babel': {
  includeExternalHelpers: true,
  includePolyfill: true,
},
```

Run the app with `ember serve` and watch it blow in a browser.

A few things to note - using `npm` instead of `yarn` fixes the issue. Downgrading all `@babel` packages to `7.17.x` also fixes the issue. I think [this](https://github.com/ant-design/pro-components/issues/5273#issuecomment-1133843012) is somewhat related.

boris-petrov:
@nicolo-ribaudo - I'm pinging you directly here because you seem to be the one who's implemented the PR (and also has been answering in a couple of issues related to this one in the Babel repo). If you're unfamiliar with Ember, `npm install -g ember-cli` should allow you to run the `ember new` and `ember serve` commands I've written above and the value in `ember-cli-build.js` should go in the hash in `new EmberApp`.

PR #534, opened 2026-07-07
Bump babel dependencies, use caret range for all (v7)

NullVoxPopuli:
bump minimum versions for babel deps -- ensure caret range is used for @babel/runtime


NullVoxPopuli:
> [!NOTE]
> I am a bot (Claude Code), commenting on behalf of @NullVoxPopuli.

## Why CI is failing, and the prior art on this problem

All failing Acceptance / Floating-dependencies jobs die on the same boot-time error before any test runs:

```
Uncaught Error: Could not find module `@babel/runtime/helpers/esm/toPropertyKey.js`
imported from `@babel/runtime/helpers/esm/defineProperty`
```

### Mechanism

When `includeExternalHelpers` is enabled, `treeForAddon` funnels `@babel/runtime/helpers/esm/*` into the app and transpiles it to AMD, registering **extensionless** module names (`@babel/runtime/helpers/esm/toPropertyKey`). Since `@babel/runtime` 7.13, the published ESM helpers import each other with explicit `.js` extensions (`import toPropertyKey from "./toPropertyKey.js"`). `amd-name-resolver` preserves the extension, so the AMD dep becomes `...toPropertyKey.js`, which is not a registered module name → the loader throws, vendor.js fails to boot, and every browser test fails.

### Version survey of `@babel/runtime` (checked the published helper files)

| Versions | Cross-helper import style | Works with this setup? |
|---|---|---|
| ≤ 7.12.18 (current pin) | `import x from "@babel/runtime/helpers/esm/x"` (extensionless) | ✅ |
| 7.13.0 – 7.13.9 | `export { default } from "../x/index.mjs"` (reverted upstream in [babel/babel#12919](https://github.com/babel/babel/pull/12919)) | ❌ |
| 7.13.10 – 7.29.7 (current) | `import x from "./x.js"` | ❌ |

So there is **no newer `@babel/runtime` that avoids this** — the fix has to be on ember-cli-babel's side.

### Prior issues / PRs discussing this

- **#384** (2021) — original report: "includeExternalHelpers option breaking with latest @babel". @rwjblue's diagnosis: *"Basically babel is including the extension in the import paths for the helpers, and our transpilation doesn't support modules at runtime including the extension."*
- **#385** — the PR that introduced the `7.12.x` pin as the mitigation.
- **#451** — "unpin @babel/runtime to see what tests fail" (still open, stalled).
- **#442** — related `externalHelpers` breakage; contains @ef4's [recommended user-side workaround](https://github.com/emberjs/ember-cli-babel/issues/442#issuecomment-2894738580) (stop using `includeExternalHelpers`, depend on `@babel/runtime` directly).
- **#445** — same error class (`regeneratorRuntime`) reported in a fresh app.
- **#519** (2025) — "Does `@babel/runtime` still need pin?" @gorner hit the identical `toPropertyKey.js` error trying `^7.26.10` and drafted a loader.js-side fix ([ember-cli/loader.js#227](https://github.com/ember-cli/loader.js/pull/227)); @ef4 declined changing loader.js/ember-cli-babel behavior broadly, as both are stable legacy code.
- **#530** — prior attempt at this same dependency bump. @ef4 relayed the tooling-team position that *"a more targeted fix is possible"* in the `treeForAddon` helpers funnel ([index.js#L271](https://github.com/emberjs/ember-cli-babel/blob/bdb4bd14853630dcc052ed2a7ed5745309eaa264/index.js#L271)): *"We can patch/adjust babel/runtime so that the imports and the defines line up correctly."*

### Path forward for this PR

Following ef4's suggestion in #530: strip the `.js` extension from relative imports while transpiling the helpers tree (a small babel plugin added only to that one `transpileTree` call in `treeForAddon`). This is scoped entirely to the helpers funnel — no changes to loader.js or to general app transpilation. A patch implementing this passes `ember test` (including the external-helpers acceptance test) and the mocha node-tests locally with `@babel/runtime@7.29.7`; it will be pushed to this PR shortly.

